PR #26, opened 2023-12-12
build: add semantic release to CI/CD

daniel-makerx:


pyproject.toml
```diff
@@ -43,6 +43,7 @@ ruff-lsp = "^0.0.45"
 pytest-cov = "^4.1.0"
 pip-audit = "^2.6.1"
 algokit-utils = "^1.3.1"
+python-semantic-release = "^8.5.0"
```
achidlow, line 46: can we add this to it's own (optional) CD group? that's a lot of deps

daniel-makerx: Done

achidlow (approved):

PR #44, opened 2023-12-18
feat: move subroutine elimination to IR stage

daniel-makerx:


src/puya/parse.py
```diff
@@ -116,6 +116,23 @@ def build_dep_graph(self) -> dict[str, Any]:
...
+def embedded_module_alias(path: Path) -> str:
+    return path.stem.strip("_")
+
+
+EMBEDDED_MODULES = {
+    p.stem: p
+    for p in sorted(
+        (
+            p
+            for p in (Path(__file__).parent / "lib_embedded").iterdir()
```
achidlow, line 128: could `.glob("*.py")` here and skip the `if`?

also I would put `Path(__file__).parent` at the module level as `_SRC_ROOT` or similar, for clarity (IMO anyway)

daniel-makerx: Done

achidlow (approved):

PR #45, opened 2023-12-18
Add support for basic 'mutation' of arc4 types.

tristanmenzel:
'mutation' in quotes as arc4 encoded data is still passed around by value meaning arrays/structs passed as parameters or even assigned to different local variables will not hold the same reference. This should serve a stopgap solution until we are able to support non-arc4 encoded arrays and structs more natively. 

See examples/arc4_types/mutation.py for usage examples. 

This PR adds support for:
 - dynamic array append and extend
 - dynamic array, static array, and struct index/property assignment
 - dynamic array pop
 - string concat

It is worth noting that arrays and structs containing variable length items (eg. string, or nested dynamic arrays) are quite inefficient to mutate. If several indexes/properties need to be updated, it can be more efficient to simply create a new array/struct.


src/puya/ir/context.py
```diff
@@ -63,7 +63,10 @@ def resolve_function_reference(
                             invocation.source_location,
                         )
                 case awst_nodes.FreeSubroutineTarget(module_name=module_name, name=func_name):
-                    func = self.module_awsts[module_name].symtable[func_name]
+                    if module_name == "_puyapy_":
```
tristanmenzel, line 66: This is required if a subroutine in the `_puyapy_` module references another subroutine inside the `_puyapy_` module. 

After moving the internal helpers into their own module - this is no longer required but maybe it will be in the future?

daniel-makerx: Yeah a bit of a quirk due to us having something in a puyapy stub and as an embedded implementation. Maybe leave it in, but add a comment to explain why it is here

src/puya/awst/nodes.py
```diff
@@ -389,13 +389,51 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
...
+@attrs.frozen
+class ArrayPop(Expression):
+    base: Expression
+
+    def accept(self, visitor: ExpressionVisitor[T]) -> T:
+        return visitor.visit_array_pop(self)
+
+
+@attrs.frozen
+class ArrayExtend(Expression):
```
daniel-makerx, line 423: I think this might be better as a statement rather than an Expression with a void wtype? As it is analogous to the AugmentedAssignment nodes

src/puya/awst_build/main.py
```diff
@@ -54,6 +54,10 @@ def transform_ast(
                 module._fullname = embedded_src.puya_module_name  # noqa: SLF001
                 module_awst = ModuleASTConverter.convert(ctx, module)
                 result[module.name] = module_awst
+            elif module_name.startswith("puya_util_"):
```
daniel-makerx, line 57: Is this still required? Embedded modules should have been caught in the previous elif block

src/puya/lib_embedded/puyapy_lib_arc4.py
```diff
@@ -0,0 +1,252 @@
...
+
+    returns: The updated bytes for the source array
+    """
+    header_cursor = start_at_index * 2
+    if start_at_index == 0:
+        tail_cursor = length * 2
+    else:
+        tail_cursor = extract_uint16(array_data, header_cursor)
+
+    for i in urange(start_at_index, length):  # noqa: B007
```
daniel-makerx, line 246: Can use _i here to avoid the ruff warning

src/puya/ir/builder.py
```diff
@@ -2304,12 +2320,757 @@ def _iterate_tuple(
...
 
+    def _invoke_puya_lib_subroutine(
+        self,
+        *,
+        method_name: str,
+        module_name: str,
+        args: list[Value],
+        source_location: SourceLocation,
+    ) -> InvokeSubroutine:
+        sub = next(
```
daniel-makerx, line 2334: With a small refactor can use the context.resolve_function_reference to resolve these instead. I will push a commit for this

src/puya/ir/context.py
daniel-makerx: Added a comment while making other changes here

src/puya/awst_build/main.py
tristanmenzel: Nope don't think it is. I'll remove it

src/puya/awst/nodes.py
tristanmenzel: I did try that however it would require a change to the expression builder to allow statements to be returned from calls. This felt like the path of least resistance.

src/puya/lib_embedded/puyapy_lib_arc4.py
tristanmenzel: Ah nice - at one point we blocked variables starting with underscore but I guess that got removed.

daniel-makerx: We still block `_` as it is treated as an Any type, but any other variable prefixed with _ doesn't cause an unused variable warning

achidlow (approved):

PR #49, opened 2023-12-20
Allow non-abi types in abi methods where the abi type can be inferred

tristanmenzel:
AVM types such as UInt64, Bytes, and bool - and tuples made up of these types can be used in ABI methods. They will be automatically encoded and decoded.


src/puya/awst/wtypes.py
```diff
@@ -464,3 +465,46 @@ def is_arc4_argument_type(wtype: WType) -> bool:
...
+    match wtype:
+        case WTuple(types=types):
+            return all(has_arc4_equivalent_type(t) for t in types)
+    return False
+
+
+def avm_to_arc4_equivalent_type(wtype: WType) -> WType:
+    if wtype is bool_wtype:
+        return arc4_bool_wtype
+    if wtype is uint64_wtype:
```
daniel-makerx, line 486: We could also support biguint_wtype here?

src/puya/ir/arc4_router.py
```diff
@@ -70,6 +71,123 @@ def btoi(bytes_arg: awst_nodes.Expression, location: SourceLocation) -> awst_nod
...
+                op_code="extract",
+                immediates=[2, 0],
+                source_location=location,
+                stack_args=[
+                    awst_nodes.ReinterpretCast(
+                        expr=bytes_arg, wtype=wtypes.bytes_wtype, source_location=location
+                    )
+                ],
+            )
+        case wtypes.ARC4Tuple(types=tuple_types):
```
daniel-makerx, line 158: Can we also handle nested tuples here easily? If not should probably disallow it in `has_arc4_equivalent_type`

src/puya/awst/wtypes.py
tristanmenzel: biguint doesn't have a fixed precision so we would have to default to uint512 I guess which wouldn't be very efficient.

achidlow: I think allowing BigUInt and mapping to UInt512 makes sense for completeness.

src/puya/awst_build/contract.py
```diff
@@ -225,13 +225,17 @@ def visit_function(
                         )
                 else:
                     for arg_wtype in arg_wtypes:
-                        if not wtypes.is_arc4_argument_type(arg_wtype):
+                        if not wtypes.is_arc4_argument_type(
```
achidlow, line 228: if you make this `not (<a> or <b>)` black should format this a bit nicer

src/puya/awst_build/eb/tuple.py
```diff
@@ -162,4 +165,39 @@ def bool_eval(self, location: SourceLocation, *, negate: bool = False) -> Expres
...
+                    right=compare_one(get_index(self.expr, i), get_index(other_expr, i)),
+                    op=BinaryBooleanOperator.and_
+                    if op == BuilderComparisonOp.eq
+                    else BinaryBooleanOperator.or_,
+                    source_location=location,
+                )
+                i += 1
+            return var_expression(result)
+
         raise TodoError(location, "TODO: tuple comparison support")
```
achidlow, line 203: probably update this TODO and make it a code error saying that operation isn't currently supported

achidlow (approved):

PR #56, opened 2024-01-12
feat: Scratch space API

tristanmenzel:
- Expose intrinsics for store and load
 - Add class option to contracts to reserve scratch space slots for user code
 - Add validation to check static usages of store and load use only reserved slots
 - Refactor IR optimisation so unrelated optimizations do not share a file
 - Include column numbers if available in logs


src/puyapy-stubs/_unsigned_builtins.pyi
```diff
@@ -10,12 +10,14 @@ from typing import TypeVar, overload
...
-    __start: int | UInt64, __stop: int | UInt64, __step: int | UInt64 = ...
-) -> Iterator[UInt64]: ...
+class urange(Iterable[UInt64]):
+    @overload
+    def __init__(self, stop: int | UInt64, /) -> None: ...
+    @overload
+    def __init__(self, start: int | UInt64, stop: int | UInt64, /) -> None: ...
+    @overload
+    def __init__(self, start: int | UInt64, stop: int | UInt64, step: int | UInt64, /) -> None: ...
+    def __iter__(self) -> Iterator[UInt64]: ...
```
achidlow, line 20: `__iter__` not required, it's inherited from `Iterable[UInt64]`

src/puyapy-stubs/_contract.pyi
```diff
@@ -1,9 +1,15 @@
...
 
-from puyapy import UInt64
+from puyapy import UInt64, urange
 
 class Contract(abc.ABC):
-    def __init_subclass__(cls, name: str | None = None, **kwargs: object): ...
+    def __init_subclass__(
+        cls,
+        name: str | None = None,
+        scratch_slots: Sequence[Union[int, urange]] = (),
```
achidlow, line 10: prefer `|` over `Union`

src/puya/awst/nodes.py
```diff
@@ -1227,6 +1249,9 @@ class ContractFragment(ModuleStatement):
     clear_program: ContractMethod | None = attrs.field()
     subroutines: Sequence[ContractMethod] = attrs.field(converter=tuple[ContractMethod, ...])
     app_state: Sequence[AppStateDefinition] = attrs.field(converter=tuple[AppStateDefinition, ...])
+    reserved_scratch_space: Sequence[ScratchSpaceReservation] = attrs.field(
```
achidlow, line 1252: make a `frozenset` instead?

src/puya/awst_build/module.py
```diff
@@ -151,8 +152,25 @@ def visit_class_def(self, cdef: mypy.nodes.ClassDef) -> None:
             # TODO: other checks above?
             else:
                 name_override: str | None = None
+                scratch_slot_reservations = list[ScratchSpaceReservation]()
                 for kw_name, kw_expr in cdef.keywords.items():
                     with self.context.log_exceptions(kw_expr):
+                        if kw_name == "scratch_slots":
```
achidlow, line 158: I think it'd be better to fold this into the match below

src/puya/awst_build/module.py
```diff
@@ -172,6 +192,56 @@ def visit_class_def(self, cdef: mypy.nodes.ClassDef) -> None:
...
+                            "Stride is not supported when specifying ranges of reserved scratch "
+                            "slots"
+                        )
+                    case _:
+                        raise CodeError("Unexpected arguments for urange")
+
+            case _:
+                raise CodeError("Unexpected value: Expected int literal or urange expression")
+
+    def _read_int_constant(self, expr: mypy.nodes.Expression) -> int:
```
achidlow, line 236: match on `IntExpr` in the above function instead?

src/puya/ir/avm_ops.py
```diff
@@ -1,4 +1,4 @@
-# AUTO GENERATED BY scripts/generate_avm_ops.py, DO NOT EDIT
+# AUTO GENERATED BY scripts\generate_avm_ops.py, DO NOT EDIT
```
achidlow, line 1: lol, we should normalise this path

src/puyapy-stubs/_unsigned_builtins.pyi
tristanmenzel: That was my impression too however I get this:
```
scratch_slots\contract.py:10: error: Cannot instantiate abstract class "urange" with abstract attribute "__iter__"  [abstract]
            for i in urange(3, 20):
```

src/puya/awst_build/module.py
tristanmenzel: We could, but `kw_value = kw_expression.accept(self)` throws an error for the scratch_slots parameter - unless we want to implement the relevant visitors so that just works?

src/puya/awst_build/module.py
tristanmenzel: Yup good call, originally I was going to support UInt64 but realised that doesn't make much sense

.pre-commit-config.yaml
```diff
@@ -7,16 +7,15 @@ repos:
         entry: poetry run black
         language: system
         minimum_pre_commit_version: 2.9.2
-        require_serial: true
         types_or: [ python, pyi ]
+        exclude: '^src/puya/(_vendor|_typeshed)/'
```
daniel-makerx, line 11: can the exclude redefinition be removed now that require_serial is false?

src/puya/awst/nodes.py
```diff
@@ -1224,6 +1226,27 @@ class AppStateDefinition(Node):
     storage_wtype: WType
 
 
+@attrs.frozen(init=False)
+class ScratchSpaceReservation(Node):
+    start_slot: int
+    stop_slot: int
+
+    def __init__(self, start_slot: int, stop_slot: int, source_location: SourceLocation) -> None:
+        if start_slot >= stop_slot:
```
daniel-makerx, line 1235: Can this be done with an attrs validator? e.g. https://github.com/algorandfoundation/puya/blob/main/src/puya/ir/models.py#L282

.pre-commit-config.yaml
tristanmenzel: yes it can

src/puya/awst/nodes.py
tristanmenzel: Um, maybe? But how would that work when we need to check two fields?

src/puya/awst_build/module.py
```diff
@@ -562,3 +583,30 @@ def visit_assignment_expr(self, o: mypy.nodes.AssignmentExpr) -> ConstantValue:
...
         return self._unsupported(expr)
+
+
+def _map_scratch_space_reservation(
+    expr: mypy.nodes.Expression, source_location: SourceLocation
+) -> Iterable[int]:
+    match expr:
+        case mypy.nodes.IntExpr(value):
+            return [value]
+        case mypy.nodes.CallExpr(callee=mypy.nodes.NameExpr(fullname=constants.URANGE), args=args):
```
achidlow, line 594: what happens here if the user introduces a `TypeAlias` for `urange`?

src/puyapy-stubs/_unsigned_builtins.pyi
achidlow: oh weird, it's declared as a Protocol in typeshed.
oh well, doesn't hurt to have it

src/puya/awst_build/module.py
achidlow: fixed by combining them

src/puya/awst/nodes.py
achidlow: I've simplified this away so no longer an issue

daniel-makerx (approved):

PR #59, opened 2024-01-17
V10 opcodes

tristanmenzel:
- Adds new v10 opcodes
 - Adds target avm version to compiler options
 - Adds check during compilation that used opcodes are available in the target avm version


src/puya/options.py
```diff
@@ -22,3 +22,4 @@ class PuyaOptions:
     debug_level: int = 0
     optimization_level: int = 0
     log_level: LogLevel = LogLevel.info
+    target_avm_version: int = 8
```
daniel-makerx, line 25: Prior to this I think we had been assuming v9, so maybe the default should be that?

Also should this result in a `#pragma version ` line in the TEAL?

tristanmenzel: yup, the line which writes the `#pragma` was hardcoded to 8 which is why I chose 8 🤷

src/puya/codegen/teal_writer.py
```diff
@@ -84,7 +84,7 @@ def emit_memory_ir_as_teal(
             annotater.header(writer)
 
         writer.new_line()
-    writer.append_line("#pragma version 8")
+    writer.append_line(f"#pragma version {context.options.target_avm_version}")
```
tristanmenzel, line 87: @daniel-makerx

src/puya/options.py
daniel-makerx: Oh I see, 8 seems fine then

src/puya/__main__.py
```diff
@@ -89,6 +89,7 @@ def main() -> None:
         type=LogLevel.from_string,
         choices=list(LogLevel),
     )
+    parser.add_argument("--target-avm-version", type=int, choices=[8, 9, 10])
```
achidlow, line 92: Needs a `default` parameter

src/puya/awst_build/intrinsic_data.py
```diff
@@ -392,6 +398,7 @@
                     allowed_types=[
                         wtypes.bytes_wtype,
                         bytes,
+                        wtypes.account_wtype,
```
achidlow, line 401: hmmm, assuming we stick with targeting multiple versions as a CLI option, how do we gate stuff like this?

src/puya/ir/avm_ops_models.py
```diff
@@ -8,6 +8,8 @@ class StackType(enum.StrEnum):
     uint64 = enum.auto()
     bytes = "[]byte"  # noqa: A003
     bytes_32 = "[32]byte"
+    bytes_64 = "[64]byte"
+    bytes_80 = "[80]byte"
```
achidlow, line 12: Hmm, looks like we don't have any special handling of these, or even `bytes_32`. This feels like langspec is leaking - @daniel-makerx can we address this higher up and also handle the `bytes_32` case?

src/puya/__main__.py
tristanmenzel: Not sure there's much value in a test as it will need to be updated on every version release and potentially won't even have a testable example in some cases. e.g. if we move to target a new version with breaking changes there wont' be any vnext opcodes you could pick to make the check fail

src/puya/ir/avm_ops_models.py
daniel-makerx: Removed the bytes_* enum values from IR models, but left in the langspec models used by scripts

achidlow (approved):

PR #60, opened 2024-01-18
[TESTING] Add job to build project on windows

tristanmenzel:


tests/test_compile.py
```diff
@@ -109,7 +109,10 @@ def compile_test_case(
         log_options = attrs.evolve(
             puya_options, out_dir=None, paths=(test_case.path.relative_to(test_case.root),)
         )
-        log_path.write_text(f"debug: {log_options}\n{compile_result.logs}", encoding="utf8")
+        log_options_str = (
```
achidlow, line 112: this shouldn't be required anymore, I've changed the output format

daniel-makerx: Removed

PR #66, opened 2024-01-25
Feature/reverse enumeration

tristanmenzel:


robdmoore (approved):

test_cases/unssa/out/contract.approval.teal
```diff
@@ -264,6 +264,8 @@ test_param_update_with_reentrant_entry_block_v2_if_body@3:
 test_param_update_with_reentrant_entry_block_v3:
         proto 0 0
 
+test_param_update_with_reentrant_entry_block_v3_block@0:
```
achidlow, line 267: hmmm, it's added an empty block?

achidlow:
looking at `sizes.txt`, a whole lot of contracts got bigger, seemingly ones without changes too?

src/puya/ir/optimize/collapse_blocks.py
```diff
@@ -119,7 +119,7 @@ def remove_empty_blocks(_context: CompileContext, subroutine: models.Subroutine)
             empty_block = block
             target = block.terminator.target
 
-            if not empty_block.predecessors and block is not subroutine.body[0]:
+            if not empty_block.predecessors and block is subroutine.body[0]:
```
achidlow, line 122: the condition was correct before, if a bit confusing.
unreachable empty blocks were deferred for removal until the `remove_unreachable_blocks` pass, except for the entry block which is always considered unreachable.
we can actually just delete this `if/continue` block entirely.

src/puya/ir/optimize/arithmetic.py
```diff
@@ -241,7 +241,13 @@ def try_simplify_arithmetic_ops(value: models.ValueProvider) -> models.ValueProv
                     case AVMOp.mul:
                         c = a_const * b_const
                     case AVMOp.div_floor:
+                        if b_const == 0:
```
daniel-makerx, line 244: We had some checks like this previously, but I think they were removed because it is possible these operations are guarded by a conditional that would prevent their evaluation at runtime

tristanmenzel: ah of course, I'll drop the `raise`

src/puya/ir/optimize/collapse_blocks.py
tristanmenzel: I see you found the actual cause of the bug and fixed it. I'll rebase and revert this change

PR #76, opened 2024-01-31
feature: Unpacking of args in calls with more than 15 args

tristanmenzel:


tests/test_arc32.py
```diff
@@ -471,6 +482,65 @@ def test_arc4_routing(
...
+    )
+    app_client = algokit_utils.ApplicationClient(algod_client, app_spec, signer=account)
+
+    # create
+    create_response = app_client.create()
+    assert create_response.confirmed_round
+
+    result = app_client.call(
+        "method_with_more_than_15_args",
+        a=1,
```
daniel-makerx, line 502: Could we pass different values for each arg just to verify things are unpacked as expected?

Maybe bytes b"a" - b"v" and concat for the result?

daniel-makerx (approved):

PR #78, opened 2024-02-01
fix: improve error output when parsing fails

daniel-makerx:


src/puya/ir/context.py
```diff
@@ -114,16 +114,11 @@ class IRBuildContextWithFallback(IRBuildContext):
...
-            self.errors.error(f"FATAL {ex!s}", location=ex.location or fallback_location)
-            crash_report(ex.location or fallback_location)
-        except Exception as ex:
-            self.errors.error(f"UNEXPECTED {ex!s}", location=fallback_location)
-            crash_report(fallback_location)
+        with log_exceptions(self.errors, fallback_location):
+            try:
+                yield
+            finally:
+                pass
```
achidlow, line 121: The with/finally here is redundant, just the yield should be the same.
Can we just return the generator?

daniel-makerx: Fixed

src/puya/compile.py
```diff
@@ -267,3 +276,20 @@ def write_contract_files(base_path: Path, compiled_contract: CompiledContract) -
...
         logger.info(f"Writing {make_path_relative_to_cwd(output_path)}")
         output_path.write_text(output_text, encoding="utf-8")
+
+
+def _log_parse_errors(ex: ParseError) -> None:
+    location: SourceLocation | None = None
+    related_errors = list[str]()
+    for error in ex.errors:
+        if error.location:  # align related error messages
+            if related_errors:
```
achidlow, line 286: can we flip this if-else?

daniel-makerx: Yep

achidlow (approved):

PR #79, opened 2024-02-02
feature: Add account/bytes ctor to arc4.Address

tristanmenzel:
Replaces the constructor for the `arc4.Address` type to take either `Bytes` or an `Account` instead of the one requiring 32 `Byte | UInt8` values inherited from `arc4.StaticArray`


src/puya/awst_build/eb/arc4/arrays.py
```diff
@@ -235,6 +240,46 @@ def __init__(self, location: SourceLocation):
...
+        args: Sequence[ExpressionBuilder | Literal],
+        arg_kinds: list[mypy.nodes.ArgKind],
+        arg_names: list[str | None],
+        location: SourceLocation,
+        original_expr: mypy.nodes.CallExpr,
+    ) -> ExpressionBuilder:
+        match args:
+            case (ExpressionBuilder() as eb,) if eb.rvalue().wtype == wtypes.account_wtype:
+                address_bytes: Expression = get_bytes_expr(eb.rvalue())
+            case (eb_or_literal,):
```
daniel-makerx, line 254: Could this handle Literal separately so the length check can be done at compile time in that case?

tristanmenzel: The optimiser should pick up the const value and optimise it away

daniel-makerx (approved):

PR #80, opened 2024-02-05
TEAL output rework & optimiser enhancements

achidlow:
The annotated output of a TEAL file has been greatly improved. The format now shows the source lines and source code above the section of associated TEAL. Extraneous information that was intended for internal development purposes has been removed (available as a separate output now with --output-memory-ir). Multi-line statements are faithfully reproduced now, so for example if an assert spans multiple lines it will appear as such in the TEAL file. Additionally, any comments on the line or on preceding lines (up to a blank line or the previous statement) will also be preserved.

Optimizer improvements:
- Implemented repeated expression elimination.
- Optimize branching on a ! condition, by swapping the branches
- Eliminated unused pure intrinsic ops.
- Inline conditional branch to an err block into an assert true/false.
- Blocks ending in a TEAL switch or match op can now have ops other than b as a fallback, in particular with err this, together with the above change, eliminates almost all standalone err blocks.
- Added a TEAL optimization pass that runs between MIR and TEAL generation, this simplifies the optimizer and also allows for crossing more virtual stack op boundaries easily.
- Added almost all cases found by the O2 brute force TEAL replacement search as special cases at O1.
- Added multiple other new optimizations to the new final TEAL optimizer.
- Minor tweaking of ARC4 embedded subroutine dynamic_array_pop_fixed_size.
- Add locals coalescing strategy as an option that it's independent of the optimization level, given that there's not necessarily one best approach.
- Significantly improved the code generated when routing >15 ABI args by removing a temporary assignment, which allows the new Repeated Expression Elimination optimizer pass to do its thing.

BREAKING CHANGES:
`--output-cssa-ir` and `--output-parallel-copies-ir` CLI options have been removed, and `--output-final-ir` is now `--output-destructured-ir` to better reflect its position in the compiler chain.

The default debug level is now 1, and only one TEAL file will be emitted for each program. To get back the previous default behaviour of only outputting an unannotated teal file, pass `-g0`.


src/puya/codegen/builder.py
```diff
@@ -329,118 +311,100 @@ def lower_block_to_teal(
...
+    if mir_output_path is not None:
+        cg_context_with_debug = attrs.evolve(
+            cg_context, options=attrs.evolve(context.options, debug_level=2)
+        )
+        mir_output = emit_memory_ir(cg_context_with_debug, subroutines)
+        mir_output_path.write_text("\n".join(mir_output), "utf8")
+
+    if context.options.optimization_level > 0:
+        for sub in subroutines:
+            # TODO: port the O2 level stuff
```
tristanmenzel, line 379: Commenting so we don't forget the todo

src/puya/ir/arc4_router.py
```diff
@@ -561,11 +561,8 @@ def map_param_wtype_to_arc4_tuple_type(wtype: wtypes.WType) -> wtypes.WType:
             [map_param_wtype_to_arc4_tuple_type(a.wtype) for a in non_transaction_args[14:]]
         )
         last_arg = load_abi_arg(15, args_overflow_wtype, location)
-        decoded_args = create_temporary_assignment(
-            value=arc4_decode(
-                last_arg, wtypes.WTuple.from_types(args_overflow_wtype.types), location
-            ),
-            location=location,
+        decoded_args = arc4_decode(
```
tristanmenzel, line 564: It does get cleaned up by the optimizer but the change here to not use a temporary variable creates some horrendous (unoptimized) teal. Perhaps we should swap to doing indexed access instead of decode => tuple item access if you're not a fan of the tempvar

src/puya/ir/context.py
```diff
@@ -16,7 +16,7 @@ class IRBuildContext(CompileContext):
     module_awsts: Mapping[str, awst_nodes.Module]
     subroutines: dict[awst_nodes.Function, Subroutine]
     embedded_funcs: Sequence[awst_nodes.Function] = attrs.field()
-    contract: awst_nodes.ContractFragment | None = attrs.field(default=None)
+    contract: awst_nodes.ContractFragment | None = None
```
tristanmenzel, line 19: Curious, is there a reason for these changes other than stylistic preference?

src/puya/ir/arc4_router.py
achidlow: I'm not too worried about the unoptimised TEAL here, especially for a niche case like >15 ABI args.

src/puya/ir/context.py
achidlow: yeah just harder to read and looks like something more complicated is going on, when it's just a default

daniel-makerx (approved):

src/puya/ir/arc4_router.py
achidlow: Oh btw, it's not just a case of "it gets cleaned up by the optimizer", it makes a huge positive difference to output when there's >15 args.

achidlow: https://github.com/algorandfoundation/puya/pull/80/commits/1fb40133ce2efd28b8c224a48abe29c6fad26a02#diff-32d307c813cea76e077972c004db204cc8c3a9387ffd7cd35cc822e0c9cb7f07

tristanmenzel: > it makes a huge positive difference to output
Only because our stack allocation still sucks :P

tristanmenzel: Pushed a [commit](https://github.com/algorandfoundation/puya/pull/80/commits/d2cf74f4b3559cae6c563420bf82ced8ffb1bdf8) to swap to indexed access. No change to the optimised output but a big change to the raw IR which should mean less work for the later stages of the compiler to do

achidlow: @tristanmenzel  nice, awesome work

PR #85, opened 2024-02-07
perf: prevent too many permutations during O2 stack optimizations

daniel-makerx:


test_cases/arc4_types/out/reference_types.O0.log
```diff
@@ -1,24 +1,24 @@
 PC Teal                 Stack
 1  <intcblock>
-4  txn Sender           0xDF1FF999B7824E21387B1AB88741BA8D95C54784CB1D91D5DF11E3ED47459CE4
```
tristanmenzel, line 3: Hmmm, maybe we should drop the logs from source control?

achidlow (approved):

test_cases/arc4_types/out/reference_types.O0.log
achidlow: the traces do come in handy sometimes, for reviewing diffs when there's loops. 
maintaining is a bit tedious though

PR #91, opened 2024-02-09
remove enums from stubs, fix enum class names in _gen

daniel-makerx:


test_cases/abi_routing/contract.py
```diff
@@ -37,10 +38,10 @@ def noop_with_uint64(self, a: arc4.UInt64) -> arc4.UInt8:
     @arc4.abimethod(
         allow_actions=[
             "NoOp",
-            "OptIn",
+            OnCompleteAction.OptIn,
             "CloseOut",
-            "UpdateApplication",
-            "DeleteApplication",
+            OnCompleteAction.UpdateApplication,
```
tristanmenzel, line 43: Any reason for the mix of strings and enum values here? Is it to prove both work?

tristanmenzel (approved):

test_cases/abi_routing/contract.py
daniel-makerx: Yeah just testing either work

PR #92, opened 2024-02-10
Op module

daniel-makerx:


examples/amm/contract.py
```diff
@@ -68,8 +64,8 @@ def bootstrap(self, seed: PaymentTransaction, a_asset: Asset, b_asset: Asset) ->
         """
         assert not self.pool_token, "application has already been bootstrapped"
         self._check_is_governor()
-        assert Global.group_size() == 2, "group size not 2"
-        assert seed.receiver == Global.current_application_address(), "receiver not app address"
+        assert op.Global.group_size() == 2, "group size not 2"
```
robdmoore, line 67: I feel like `Global` and `Transaction` shouldn't be in `op`, but should be standalone?

robdmoore: Or alternatively, maybe they could be lit up into `self` via the base `Contract` class as convenient aliases?

daniel-makerx: Yeah we plan to make some more changes to make here, for Global and Transaction, any values that don't change or take arguments will be converted to properties.

I think Global and Transaction should remain in `puyapy.op` as they represent a 1:1 mapping to the underlying teal ops. But it would also think it is reasonable to also include them in the `puyapy` module as aliases, since they are commonly used.

Having them available via `self` is tempting from a discoverability point of view, but the main issue is they are contextual info of the current transaction rather than contract state. Currently all other things on `self` relate directly to state values, but keen to hear thoughts on this.

robdmoore:
Sounds good!--On 12 Feb 2024, at 6:04 pm, Daniel McGregor ***@***.***> wrote:﻿
@daniel-makerx commented on this pull request.



In examples/amm/contract.py:
> @@ -68,8 +64,8 @@ def bootstrap(self, seed: PaymentTransaction, a_asset: Asset, b_asset: Asset) ->
         """
         assert not self.pool_token, "application has already been bootstrapped"
         self._check_is_governor()
-        assert Global.group_size() == 2, "group size not 2"
-        assert seed.receiver == Global.current_application_address(), "receiver not app address"
+        assert op.Global.group_size() == 2, "group size not 2"

Yeah we plan to make some more changes to make here, for Global and Transaction, any values that don't change or take arguments will be converted to properties.
I think Global and Transaction should remain in puyapy.op as they represent a 1:1 mapping to the underlying teal ops. But it would also think it is reasonable to also include them in the puyapy module as aliases, since they are commonly used.
Having them available via self is tempting from a discoverability point of view, but the main issue is they are contextual info of the current transaction rather than contract state. Currently all other things on self relate directly to state values, but keen to hear thoughts on this.

tristanmenzel (approved):

src/puyapy-stubs/op.pyi
```diff
@@ -2142,56 +2142,140 @@ class EllipticCurve:
 class Global:
     """Functions for the op: `global`"""
 
-    @staticmethod
-    def min_txn_fee() -> UInt64:
-        """
-        global field F
+    min_txn_fee: UInt64 = ...
```
achidlow, line 2145: Can we annotate all these with `ClassVar`?

src/puyapy-stubs/op.pyi
```diff
@@ -2200,1458 +2284,826 @@ class Global:
...
-        """
-    @staticmethod
-    def latest_timestamp() -> UInt64:
-        """
-        global field F
+class InnerTransaction:
+    """Functions for the ops: `itxn`, `itxnas`"""
 
-        Groups: Loading Values
+    sender: Account = ...
```
achidlow, line 2347: These maybe shouldn't be properties? They're not constants.

src/puyapy-stubs/op.pyi
achidlow: Also, I don't suppose there's any way we can mark these as readonly... I know you mentioned that having this as a type and then an instance so that properties are usable didn't work with the doc generation :(

daniel-makerx: Done

src/puyapy-stubs/op.pyi
daniel-makerx: Good point, done

src/puya/awst_build/eb/log.py
```diff
@@ -0,0 +1,123 @@
...
+        location: SourceLocation,
+        original_expr: mypy.nodes.CallExpr,
+    ) -> ExpressionBuilder:
+        args_ = list(args)
+        try:
+            sep_index = arg_names.index("sep")
+        except ValueError:
+            sep: Expression = BytesConstant(value=b"", source_location=location)
+        else:
+            sep = expect_operand_wtype(args_.pop(sep_index), wtypes.bytes_wtype)
```
achidlow, line 47: this will need to handle `str` literals, since it's allowed in your stub definition

src/puya/awst_build/eb/log.py
```diff
@@ -0,0 +1,123 @@
...
+            IntrinsicCall(
+                op_code="log",
+                wtype=wtypes.void_wtype,
+                stack_args=[log_value],
+                source_location=location,
+            )
+        )
+
+
+def itob(expr: Expression, loc: SourceLocation) -> Expression:
```
achidlow, line 108: these should preferably be underscore prefixed, so we don't get them as suggested imports etc

src/puya/awst_build/eb/log.py
daniel-makerx: Done

achidlow (approved):

PR #95, opened 2024-02-13
refactor: separate IR builder into multiple modules

daniel-makerx:


daniel-makerx:
> Nice! I assume the lack of output changes is proof there are no regressions

Yep, no change to output with this

PR #96, opened 2024-02-14
fix: when simplifying a control op by inlining a block then ensure successor phi arguments are also updated

daniel-makerx:


examples/sizes.txt
```diff
@@ -26,6 +26,7 @@ chained_assignment                   83      83      83
 conditional_execution                380     344     344     
 conditional_expressions              229     181     181     
 contains                             155     156     156     
+control_op_simplification            44      48      48
```
daniel-makerx, line 29: @achidlow Note the increase in size here, if the optimisation is disabled, then the size goes down to 40 rather than up. However it seems the increase is due to additional copies introduced during SSA destructuring.

src/puya/ir/optimize/control_op_simplification.py
```diff
@@ -190,13 +191,16 @@ def remove_target(parent: models.BasicBlock, to_remove: models.BasicBlock) -> No
                 if default_block not in block.terminator.targets():
                     remove_target(block, default_block)
                 # add this block as a predecessor to any new targets
+                # and update relevant phi args in successor
                 for succ in unique(block.successors):
                     if block not in succ.predecessors:
```
daniel-makerx, line 196: If `block` is in `succ.predecessors` what should happen with Phi arguments in `succ`? Perhaps this optimization should only occur if all successor phi argument values for `block` match  phi argument values for `default_block`?

achidlow: Yeah, for now though we can also just check for intersection between targets of `default_block` and `block` and not proceed if there is one.

examples/sizes.txt
achidlow: Hmmm, might need to look at this in another PR.

achidlow (approved):

PR #97, opened 2024-02-15
Box Storage API

tristanmenzel:


scripts/generate_stubs.py
```diff
@@ -102,7 +102,7 @@ def includes_op(self, op: str) -> bool:
         },
     ),
     GroupedOpCodes(
-        name="Box",
+        name="BoxStatic",
```
tristanmenzel, line 105: This might not be necessary now that we're splitting the ops into a separate module

examples/voting/voting.py
```diff
@@ -234,20 +236,14 @@ def allowed_to_vote(self, signature: Bytes) -> bool:
             self.snapshot_public_key,
         )
 
+    @subroutine
+    def get_vote_from_box(self, index: UInt64) -> UInt64:
+        return btoi(self.tallies.extract_bytes(index, VOTE_COUNT_BYTES))
```
robdmoore, line 241: Can this be a .decode() rather than btoi and extract_bytes?

examples/voting/voting.py
```diff
@@ -107,7 +110,7 @@ def bootstrap(self, fund_min_bal_req: PaymentTransaction) -> None:
         assert (
             fund_min_bal_req.amount == min_balance_req
         ), "Payment must be for the exact min balance requirement"
-        assert Box.create(TALLY_BOX_KEY, tally_box_size)
+        self.tallies.value = bzero(tally_box_size)
```
robdmoore, line 113: What’s bzero?

tristanmenzel: creates a byte array of length N (all zeros)

examples/voting/voting.py
tristanmenzel: if we represented `tallies` as a `arc4.DynamicArray[arc4.UInt64]` then we could do something like `return self.tallies[index].decode()` but then we'd have the overhead storing a length header in box storage unnecessarily. We can't use `arc4.StaticArray` as the length is not known at compile time.

tristanmenzel: we could add some new methods to the `Bytes` type such as `my_bytes.extract_uint64(index)`

achidlow:
Some general comments:
- Boxes aren't tied to the lifetime of the contract, so don't necessarily need to be done through `self.`
- Let's come up with a robust single box abstraction before we tackle some sort of mapping - ideally I think we want to offload some stuff like this out of the core compiler somehow, maybe. We should discuss it at least.
- The `Box` API feels a bit confused with value property there, and the booleanness being on existence, and also some top level value mutating methods.
- In #98 there's a `BytesBacked` protocol. We can restrict the `_T` on `Box` to either `Bytes | BytesBacked` I think.

stubs/algopy-stubs/_box.pyi
```diff
@@ -0,0 +1,193 @@
...
+    Box abstracts the reading and writing of a single value to a single statically keyed box.
+    The box size will be reconfigured dynamically to fit the size of the value being assigned to it.
+    """
+
+    def __init__(
+        self,
+        type_: type[_TContent],
+        /,
+        *,
+        key: Bytes | bytes = ...,
```
daniel-makerx, line 20: Based on the description and implementation,  Bytes shouldn't be here?

stubs/algopy-stubs/_box.pyi
```diff
@@ -0,0 +1,193 @@
...
+    def __bool__(self) -> bool:
+        """Returns True if the box has a value set, regardless of the truthiness of that value"""
+    @typing.overload
+    def create(self, value: Bytes | bytes) -> None:
+        """
+        Creates a box with the specified value. Fails if the box already exists with a
+        different size to `len(value)`
+        """
+    @typing.overload
+    def create(self, size: UInt64 | int) -> bool:
```
daniel-makerx, line 78: Should this be it's own method rather than an overload? Since in the bytes overload it stores the passed value, but in this uint64 overload it is just allocating the space

src/puya/awst_build/eb/box.py
```diff
@@ -0,0 +1,538 @@
...
+    box_key: Expression = expect_operand_wtype(key, box_dec.key_wtype)
+
+    if wtypes.is_uint64_on_stack(box_key.wtype):
+        box_key = intrinsic_factory.itob(
+            ReinterpretCast(expr=box_key, wtype=wtypes.uint64_wtype, source_location=location),
+            location,
+        )
+    if box_dec.key_prefix:
+        box_key = (
+            var_expression(box_key)
```
daniel-makerx, line 78: should this be the other way around? i.e. prefix first

src/puya/awst_build/eb/box/_util.py
```diff
@@ -0,0 +1,77 @@
...
+                begin_index_expr,
+                UInt64Constant(value=1, source_location=location),
+            ],
+            source_location=location,
+            wtype=wtypes.bytes_wtype,
+        )
+    )
+
+
+def slice_box_bytes(
```
tristanmenzel, line 45: This produces some ugly awst, but I moved away from using SliceExpression as it it was expecting `base` to be a `bytes` type which it isn't always (at last not directly). I wonder if it's worth adding `BoxSlice` as a node to clean this up.  🤔

achidlow: The supported types in the AWST validation are purely for early validation, if it makes sense to add a new supported type here, we can use SliceExpression.

That said, I need some more time to review this PR. I think some of the abstractions might be leaking a bit, and especially with the planned TypeScript work starting shortly, it's important to get them right. So that might change what we do here too.

achidlow:
Closing this PR, since context has expanded and current reviews are not necessarily relevant

PR #98, opened 2024-02-15
feat: add API for creating and submitting inner transactions

daniel-makerx:


docs/transactions.md
```diff
@@ -0,0 +1,166 @@
...
+
+Inner transactions are powerful, but currently do have some restrictions in how they are used.
+
+* Inner transaction objects cannot be passed to subroutines
+  ```{note}
+  Individual fields can be passed to subroutines
+  ```
+* Inner transaction parameters cannot be reassigned without a `.copy()`
+* Inner transactions cannot be reassigned
+* Inner transactions array values cannot be accessed if there is a subsequent inner transaction submitted.
```
tristanmenzel, line 162: I'm not quite sure what this means - maybe an example is in order

examples/auction/contract.py
```diff
@@ -82,45 +83,40 @@ def bid(self, pay: PaymentTransaction) -> None:
...
+            receiver=op.Transaction.sender,
+        ).submit()
 
         self.claimable_amount[op.Transaction.sender] = original_amount - amount
 
     @arc4.abimethod
     def claim_asset(self, asset: Asset) -> None:
         assert op.Global.latest_timestamp > self.auction_end, "auction has not ended"
-
+        previous_bidder = self.previous_bidder
```
tristanmenzel, line 103: Can we do this in the compiler rather than expecting the user to know it needs to be copied to a local? 🤔

tristanmenzel: or is the copy to local just an optimization to prevent global state reads? is that something we need to do?

src/puya/awst/wtypes.py
```diff
@@ -115,51 +115,57 @@ def is_valid_utf8_literal(value: object) -> typing.TypeGuard[str]:
...
-    stub_name=constants.CLS_ASSET_FREEZE_TRANSACTION_ALIAS,
-)
 
-application_call_wtype: typing.Final = WTransaction(
-    transaction_type=constants.TransactionType.appl,
-    name="application_call",
-    stub_name=constants.CLS_APPLICATION_CALL_TRANSACTION_ALIAS,
-)
+@attrs.define
+class WInnerTransaction(WType):
```
tristanmenzel, line 156: winner winner chicken dinner!

tristanmenzel (approved):

examples/auction/contract.py
daniel-makerx: Yeah this change isn't required for the inner transaction to work, was just a minor optimisation to prevent two reads (and associated asserts). But I'll remove the changes from this PR, as it does make it seem like that.

There is probably an optimisation here that could be done to treat duplicate state reads as a repeated sub expression if there is no write between them

docs/transactions.md
daniel-makerx: Added examples for all of these scenarios

PR #102, opened 2024-02-19
feat: rename some op code types to more closely reflect the underlying op code

daniel-makerx:


examples/amm/contract.py
```diff
@@ -32,7 +32,7 @@ def __init__(self) -> None:
         # The asset id of asset B
         self.asset_b = Asset(0)
         # The current governor of this contract, allowed to do admin type actions
-        self.governor = op.Transaction.sender
+        self.governor = op.Txn.sender
```
robdmoore, line 35: Do we have good intellisense docs for this?

robdmoore: Also, did we add an alias for this in the route?

daniel-makerx: Currently we have doc that is generated from the langspec. But I think at some point we need to switch to handwritten doc.

Will add an alias for `Txn` and `Global` to puyapy as part of this PR

achidlow (approved):

PR #106, opened 2024-02-22
build: Use module paths for mypy pre-commit instead of filenames to w…

tristanmenzel:
…ork around command length issue on windows

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,8 @@ repos:
       - id: mypy
         name: mypy
         description: '`mypy` will check Python types for correctness'
-        entry: poetry run mypy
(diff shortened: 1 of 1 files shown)


achidlow:
Doesn't this mean every single file will be type checked on every pre-commit, massively slowing it down? I'd prefer if we found another way around the line length issue - is there any prior art we can leverage?

tristanmenzel:
We could maybe wait for this to be resolved as I suspect it might be the cause: https://github.com/pre-commit/pre-commit/issues/3124 

@achidlow

achidlow:
We need to get to the bottom of the issue, as per the linked issue in `pre-commit` - @tristanmenzel can you attach a debugger and see what's going on on Windows?

PR #107, opened 2024-02-22
feat: allow passing mutable ARC4 types for mutation

tristanmenzel:
i.e. passing an `arc4.Struct` or `arc4.Static/DynamicArray` to another function is allowed, without a `.copy()` being required, since the "pass by reference" semantics will be correctly respected.

Note that this doesn't account for aliasing - in the case of nested ARC4 types, only the outermost can be passed this way.


test_cases/arc4_types/mutable_params.py
```diff
@@ -0,0 +1,108 @@
...
+        array[0] = UInt8(10)
+        return copy
+
+    @subroutine
+    def other_routine_3(self, arrays: tuple[TestArray, TestArray, TestArray]) -> None:
+        # This doesn't mutate the params
+        for array in arrays:
+            array[0] = UInt8(99)
+
+        # This also doesn't work
```
tristanmenzel, line 96: This fails because we don't allow an assignment to a tuple item (which is how array mutations are currently expanded) we could add an edge case to allow it for mutable arc4 tuple values

test_cases/arc4_types/mutable_params.py
```diff
@@ -0,0 +1,108 @@
...
+        arrays[0][0] = UInt8(99)
+        arrays[1][0] = UInt8(99)
+        arrays[2][0] = UInt8(99)
+
+        # And this doesn't work
+        (one, two, three) = arrays
```
tristanmenzel, line 102: This doesn't work because `one` is a different register to `arrays.0`. Should we force a copy here? if so - where does it go?

test_cases/arc4_types/mutable_params.py
```diff
@@ -0,0 +1,108 @@
...
+    @subroutine
+    def other_routine_2(self, array: TestArray) -> TestArray:
+        copy = array.copy()
+        array[0] = UInt8(10)
+        return copy
+
+    @subroutine
+    def other_routine_3(self, arrays: tuple[TestArray, TestArray, TestArray]) -> None:
+        # This doesn't mutate the params
+        for array in arrays:
```
tristanmenzel, line 93: The `array` item_var doesn't match `arrays.0` so updates are not propagated

test_cases/arc4_types/mutable_params.py
```diff
@@ -0,0 +1,108 @@
...
+        # Pass to subroutine with copy
+        self.other_routine(my_array_copy.copy(), my_struct_copy.copy())
+
+        assert my_array_copy[1] == UInt8(2), "my_array_copy should not be mutated by the subroutine"
+
+        assert my_struct_copy.s_val_1 == String("Happy"), "my_struct_copy should not be mutated by the subroutine"
+
+        my_array_copy_2 = my_array_copy.copy()
+
+        my_array_copy_2 = self.other_routine_2(my_array_copy_2)
```
tristanmenzel, line 65: Verify that explicit return values overrule the implicitly returned ones

examples/sizes.txt
```diff
@@ -61,5 +62,5 @@ tuple_support                        438     290     290
 unary                                130     92      92      
 undefined_phi_args/baddie            346     346     311     
 unssa                                442     371     371     
-voting/voting                        1605    1505    1505    
+voting/voting                        1624    1523    1523
```
robdmoore, line 65: Why the increase?

.pre-commit-config.yaml
```diff
@@ -22,7 +22,8 @@ repos:
       - id: mypy
         name: mypy
         description: '`mypy` will check Python types for correctness'
-        entry: poetry run mypy
+        entry: poetry run mypy src/puya src/puyapy-stubs test_cases tests examples
```
achidlow, line 25: This looks like #106 accidentally made its way here? We definitely don't want to do this because it'll force checking everything on every commit.

src/puya/awst_build/validation/arc4_copy.py
```diff
@@ -23,7 +23,7 @@ def visit_assignment_statement(self, statement: awst_nodes.AssignmentStatement)
 
     def _check_for_arc4_copy(self, expr: awst_nodes.Expression) -> None:
         match expr.wtype:
-            case wtypes.ARC4Array() | wtypes.ARC4Struct():
+            case wtypes.ARC4Array(immutable=False) | wtypes.ARC4Struct(immutable=False):
```
achidlow, line 26: looks like the `WType`s for these need updating to actually be made to be `immutable=False`?

src/puya/awst_build/validation/arc4_copy.py
```diff
@@ -44,8 +44,3 @@ def _check_for_arc4_copy(self, expr: awst_nodes.Expression) -> None:
     def visit_assignment_expression(self, expr: awst_nodes.AssignmentExpression) -> None:
         super().visit_assignment_expression(expr)
         self._check_for_arc4_copy(expr.value)
-
-    def visit_subroutine_call_expression(self, expr: awst_nodes.SubroutineCallExpression) -> None:
```
achidlow, line 48: Wouldn't this still need to be in there, in order to check nested items, e.g. an arc4 array inside a regular tuple?

achidlow: Or, looks like maybe you got that case working? But we'd still need to check if say they passed a reference to a nested arc4.Struct or some such, right?

test_cases/arc4_types/mutable_params.py
achidlow: This one is a bit of a pickle, but not specific to subroutine arguments as you've noted.
Adding  a `.copy()` before iteration (which isn't even available on tuple, lol) also wouldn't solve it, because the expected semantics there would be that the collection is copied, but references inside are maintained.
This seems more like a bug in our iteration of tuples with mutable items inside?

test_cases/arc4_types/mutable_params.py
achidlow: That's a 🥒
Can we just disallow this entirely for now?

test_cases/arc4_types/mutable_params.py
achidlow: This is effectively the same problem as the iteration - unpacking like this actually performs an iteration in real Python

src/puya/ir/main.py
```diff
@@ -146,10 +173,29 @@ def _expand_tuple_params(args: Sequence[awst_nodes.SubroutineArgument]) -> Itera
...
+
+    Anything else would require further analysis, so err on the side of caution and include include
+    the implicit returns.
+    """
+    if not isinstance(func, awst_nodes.ContractMethod) or not func.abimethod_config:
+        return True
+    return any(
+        f
+        for f in ctx.function_call_sites.get(func, [])
+        if isinstance(f, awst_nodes.ContractMethod) and f.name != "approval_program"
```
achidlow, line 194: hmmm the string comparison here is a bit tenuous but I can see why you've done it that way

examples/sizes.txt
achidlow: One of the subroutines takes a mutable type as an argument.
It's never used as such, the detection of whether to generate the implicit returns is simply "is this method called from anywhere expect an ARC4 Router" is MVP and can definitely be improved.
But it's a small price to pay for the ability to have mutable arguments that function as expected.

.pre-commit-config.yaml
tristanmenzel: Without it, `pre-commit run --all` doesn't work on windows, but perhaps we can work around it by adding a poetry script that does the same thing

src/puya/awst_build/validation/arc4_copy.py
tristanmenzel: They already are set to False

src/puya/awst_build/validation/arc4_copy.py
tristanmenzel:
```py
    @subroutine
    def check_copy_required(self) -> None:
        d_array = DynamicArray(String("Test"))

        nested_in_array = DynamicArray(d_array)

        nested_in_struct = StructWithArray(d_array=nested_in_array[0])

        assert d_array == nested_in_struct.d_array
```

Are we thinking these scenarios should require a copy? They wouldn't be under subroutine calls though

src/puya/ir/main.py
tristanmenzel: is there a more robust way to identify an approval_program other than by name? I guess we need to know if the approval_program was generated by puya, or manually too.

src/puya/awst_build/validation/arc4_copy.py
achidlow: Ah right - in the factories, but not in the types themselves. Sorry, not sure how I missed that.

But it would be safer if we add
`immutable: bool = attrs.field(default=False, init=False)`
to the base wtype of the arrays `wtypes.ARC4Array`.

src/puya/awst_build/validation/arc4_copy.py
achidlow: Specifically for this PR I was thinking `some_func(nested_in_struct.d_array)` should still error.

src/puya/ir/main.py
achidlow: Yeah it's not particularly easy currently especially with folding, we might need to do a slight refactor there to expose that neatly, I'd be okay if we merged this as is - maybe add a `REFACTOR` comment :P

src/puya/awst_build/validation/arc4_copy.py
tristanmenzel: I'll address both

achidlow (approved):

PR #111, opened 2024-03-01
Local/GlobalState custom keys & descriptions

achidlow:
Add the ability to specify a custom `key` and/or provide a `description` for `LocalState` and `GlobalState` objects. The resolution of a member name to the key has thus been deferred to the IR phase. To support this, additional nodes have been introduced for dealing with state keys by member name, which has also allowed for an approach that significantly improves the generated code in certain cases.

Solve the issue of requiring module level constructs (such as constants) to come before their usage

scratch_slots reservation fixes:
- allow module constants args, and in args to `urange` within
- handle negatives
- check for third arg to `urange` being non-zero to avoid crash
- allow just a single `urange` call
- allow list exprs
- add tests for the above and tests for failure cases

Fixes a bug where types declared in other modules wouldn't work on import.

Check file encodings and warn if it's not UTF-8.

Fix potential issue with TemporaryAssignmentExpressionBuilder not being picked up as a value.

Add "unknown" bytes encoding for raw bytes. default to hex output, but we can choose a better encoding when e.g. combining constants in optimisation (will be addressed in another PR).

Don't generate code/calls to an empty `__init__` function (todo in another PR: remove all empty function calls, and remove unreferenced functions after optimisation).


examples/amm/out/ConstantProductAMM.approval.teal
```diff
@@ -252,7 +252,7 @@ _check_is_governor:
     int 0
     byte "governor"
     app_global_get_ex
-    assert // check value exists
+    assert // check governor exists
```
robdmoore, line 255: Nice!

src/puya/awst_build/eb/app_account_state.py
```diff
@@ -115,31 +121,13 @@ def call(
...
-                value=state_def.key, encoding=state_def.key_encoding, source_location=location
-            ),
-        ],
-        wtype=wtypes.WTuple.from_types((state_def.storage_wtype, wtypes.bool_wtype)),
-    )
-    return app_local_get_ex
-
-
 def _validated_index_expr(index: ExpressionBuilder | Literal) -> Expression:
     # TODO: FIXME
```
tristanmenzel, line 132: 🤔

src/puya/awst_build/eb/temporary_assignment.py
```diff
@@ -11,14 +12,20 @@
 from puya.parse import SourceLocation
 
 
+# TODO: this should probably extend ValueProxyExpressionBuilder
```
tristanmenzel, line 15: Can we do this in this PR, or create an issue to track it rather than adding more todo comments

src/puya/context.py
```diff
@@ -33,7 +33,9 @@ def try_get_source(self, location: SourceLocation | None) -> SourceMeta:
             start_line = end_line = location.line
             if location.end_line:
                 end_line = location.end_line
-            src_content = source_lines[start_line - 1 : end_line]  # location.lines is one-indexed
+            src_content = list(
+                source_lines[start_line - 1 : end_line]  # location.lines is one-indexed
```
tristanmenzel, line 37: Would it make sense to just convert the line index to 0-based when we create the `SourceLocation` rather than having to remember this 'gotcha' everywhere we use it?

src/puya/ir/builder/state.py
```diff
@@ -0,0 +1,217 @@
...
+    context.block_builder.add(
+        Intrinsic(
+            op=AVMOp.assert_,
+            args=[did_exist_tmp],
+            comment=f"check {expr.field_name} exists",
+            source_location=expr.source_location,
+        )
+    )
+
+    return value_tmp
```
tristanmenzel, line 56: This code is a duplicate of `visit_checked_maybe`. Is it worth extracting a helper? 🤔

tristanmenzel: also duplicated below

src/puya/ir/main.py
```diff
@@ -276,41 +258,45 @@ def fold_state_and_special_methods(
...
             result.clear_program = c.clear_program
-        for state in c.app_state:
+        for state in c.app_state.values():
             translated = ContractState(
                 name=state.member_name,
                 source_location=state.source_location,
                 key=state.key,
                 storage_type=wtype_to_storage_type(state.storage_wtype),
-                description=None,  # TODO, have some way to provide this
+                description=state.description,
```
tristanmenzel, line 275: 🎉

src/puya/parse.py
```diff
@@ -248,21 +256,78 @@ def parse_and_typecheck(paths: Sequence[Path], mypy_options: mypy.options.Option
...
+        return
+    # below is based on mypy/util.py:decode_python_encoding
+    # check for BOM UTF-8 encoding
+    if source.startswith(b"\xef\xbb\xbf"):
+        return
+    # otherwise look at first two lines and check if PEP-263 coding is present
+    encoding, _ = mypy.util.find_python_encoding(source)
+    if encoding != "utf8":
+        logger.warning(
+            f"UH OH SPAGHETTI-O's,"
```
tristanmenzel, line 324: This needs to be its own base `Error`

tristanmenzel (approved):
LGTM

src/puya/awst_build/eb/temporary_assignment.py
achidlow: I've fixed the actual problem issues in this PR - namely that some places were checking for instances of `ValueProxyExpressionBuilder` instead of looking at `value_type`, and also made this class properly implement said property. 
I don't want to make this PR any bigger by doing this additional refactoring, and I don't think it's worthy of an issue, so I'm just gonna change the comment from a TODO to a REFACTOR :P

src/puya/context.py
achidlow: I considered that, but the only time it matters is when matching up a `SourceLocation` to source code, which is done only here. Having the location info be different from the Python AST could also be a different kind of gotcha. The balance may swing towards zero indexed when we start working on additional languages, if the majority of those return zero indexed lines.

src/puya/parse.py
achidlow: It's a warning because we correctly handle the encoding in terms of reading in the code, but any unicode values (e.g. attribute names or string literals) that also get put onto the chain (such as through state keys) _might_ have different encoded bytes on-disk vs on-chain.
The spaghetti was left in at request of some folks, the message regarding compound types is getting improved/removed in another PR (#104)

src/puya/ir/builder/state.py
achidlow: Done some refactoring in the state file to extract some patterns here.
The code in `visit_checked_maybe` did fit the pattern but it didn't need to, so that's also been refactored.

src/puya/parse.py
tristanmenzel: Sorry, I forgot to add the silly face emoji 😝

it was a joke… though I do think we should be able to raise an UhOhSpaghettiOError 😀

src/puya/awst_build/eb/app_account_state.py
achidlow: Wasn't part of this PR but I've now simplified this anyway.

src/puya/parse.py
achidlow: Spaghetti-o's are srs bznss m8 😤

PR #114, opened 2024-03-05
feat: Template variables

tristanmenzel:
Fixes [#29](https://github.com/algorandfoundation/algokit-internal/issues/29)


src/puya/ir/models.py
```diff
@@ -223,6 +223,15 @@ def accept(self, visitor: IRVisitor[T]) -> T:
         return visitor.visit_biguint_constant(self)
 
 
+@attrs.frozen
+class DeployVar(Value):
```
tristanmenzel, line 227: I opted for introducing a new model instead of changing the existing integer/bytes constant types as there's a lot of optimization code which works on these constants which wouldn't apply to deploy time vars

src/puyapy-stubs/_deploy_time_variables.pyi
```diff
@@ -0,0 +1,7 @@
+from puyapy import Bytes, UInt64
+
+class UInt64DeployVar(UInt64):
```
tristanmenzel, line 3: These could maybe just be functions rather than classes?

test_cases/deploy_time_variables/contract.py
```diff
@@ -0,0 +1,11 @@
+from puyapy import BytesDeployVar, UInt64, UInt64DeployVar, arc4, Bytes
+
+
+class DeployTimeVariablesContract(arc4.ARC4Contract):
```
tristanmenzel, line 4: Once we're happy with the awst and below implementation,

I'm envisioning class options to specify updatable/deletable with a value of `true | false | 'from_deploy_var' | None` and we will auto generate the code. `None` would mean no auto generated code and the user could write an explicit implementation. Might be worth exporting the `UPDATABLE` and `DELETABLE` variable names as constants in case someone wants an explicit implementation still governed by the variables the app clients provide

src/puya/mir/models.py
```diff
@@ -53,6 +53,30 @@ def __str__(self) -> str:
...
+    atype: AVMType
+
+    def accept(self, visitor: MIRVisitor[_T]) -> _T:
+        return visitor.visit_push_deploy_var(self)
+
+    @property
+    def op_code(self) -> str:
+        match self.atype:
+            case AVMType.bytes:
+                return "pushbytes"
```
tristanmenzel, line 68: I used `pushbytes` and `pushint` here to match beaker - but I'm not sure there's a reason not to just use `bytes` and `int`

test_cases/deploy_time_variables/out/DeployTimeVariablesContract.approval.teal
```diff
@@ -0,0 +1,87 @@
...
+
+// test_cases.deploy_time_variables.contract.DeployTimeVariablesContract.get_uint() -> uint64:
+get_uint:
+    // deploy_time_variables/contract.py:9-10
+    // @arc4.abimethod()
+    // def get_uint(self) -> UInt64:
+    proto 0 1
+    // deploy_time_variables/contract.py:11
+    // return UInt64DeployVar("SOME_UINT")
+    pushint TMPL_SOME_UINT
```
tristanmenzel, line 86: Beaker includes a comment with the variable name after it. We probably don't need that if we're including source comments?

```
pushint TMPL_DELETABLE // TMPL_DELETABLE
```

src/puya/awst/nodes.py
```diff
@@ -370,6 +370,15 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
         return visitor.visit_bytes_constant(self)
 
 
+@attrs.frozen
+class DeployVar(Expression):
```
daniel-makerx, line 374: Do we want to stick with the Template nomenclature (e.g `TemplateVar`)? Deployment is one use case for these, but not the only one?

src/puya/ir/models.py
daniel-makerx: That makes sense, but also got me thinking that it might be a nice addition to allow specifying the values for these variables during compilation, a bit like `-D` style macros in c compiler's. If those values are provided then these could become actual constants and would allow optimisation

src/puya/mir/models.py
daniel-makerx: I believe the reason beaker did this, is because it would allow it to do on chain substitution of template variables in logic signatures. If the `bytes` or `int` pseudo ops were used the values that needed to be replaced would end up in the int and bytes constant blocks. 

I think the current plan is to not support these use cases in puyapy, is that correct @achidlow?

src/puyapy-stubs/_deploy_time_variables.pyi
daniel-makerx: Functions might be better, as we want to avoid non protocol inheritance of the stub types

daniel-makerx:
Something to consider with this work is whether using the new `#define` directive in TEAL is helpful here

e.g. https://github.com/algorand/go-algorand/blob/298e53a6c07cd2b39428afafb75b5fc502836645/data/transactions/logic/assembler_test.go#L3295

src/puya/ir/models.py
achidlow: @daniel-makerx  the `-D` feature probably works better if/when we allow `os.getenv` at the module level.

@tristanmenzel can we rename this to `TemplateVar`?

src/puya/mir/models.py
```diff
@@ -53,6 +53,30 @@ def __str__(self) -> str:
...
+            case AVMType.uint64:
+                return "pushint"
+            case _:
+                raise InternalError(
```
achidlow, line 72: Maybe move this validation to an `attrs` validator? or even, since it's fixed, make this an `attrs` `.default` method and do the validation at the same time as op code selection, but it'll be done on construction at least.

src/puyapy-stubs/_deploy_time_variables.pyi
achidlow: hmmmmm interesting, yeah I think a function might be the way to go here, if it all works the same with the expression builders (which it should I assume?)

test_cases/deploy_time_variables/contract.py
achidlow: I think this can just be a mixin contract, rather than couple it tightly into the compiler as a class option?

tristanmenzel: Sounds good - where would we put this contract? I assume it wouldn't go in the stubs (since it would have an actual implementation? I've added constants for `TMPL_UPDATABLE` and `TMPL_DELETABLE` but these also required some compiler gymnastics to resolve in the compiler because they are only stubs.

scripts/compile_all_examples.py
```diff
@@ -105,12 +108,24 @@ class CompilationResult:
...
 def get_program_size(path: Path) -> int:
     try:
-        program = algokit_utils.Program(path.read_text("utf-8"), ALGOD_CLIENT)
+        program = algokit_utils.Program(replace_templates(path.read_text("utf-8")), ALGOD_CLIENT)
         return len(program.raw_binary)
     except Exception as e:
         raise Exception(f"Error compiling teal application: {path}") from e
 
 
+def replace_templates(
```
tristanmenzel, line 117: Have to replace templates with something valid before compiling with algod or it goes 💥

test_cases/deploy_time_variables/contract.py
achidlow: I'm thinking it would be in another repo / library entirely perhaps? `fpuyapy-utils` :P
We could shove `itoa` in there too haha

achidlow (approved):

PR #117, opened 2024-03-07
feat: spike out what unit testing puyapy contracts could look like

neilcampbell:


src/puyapy_mocks/__init__.py
```diff
@@ -0,0 +1,22 @@
+# ruff: noqa
+# types are in the correct dependency order
+
+from puyapy_mocks._primatives import *
+from puyapy_mocks._reference import *
+from puyapy_mocks._contract import *
+from puyapy_mocks._hints import *
+from puyapy_mocks._util import *
+from puyapy_mocks import (
+    arc4 as arc4,
```
aorumbayev, line 10: given that aliased name is identical, wouldn't

```py
from puyapy_mocks import (
    arc4,
    gtxn,
    itxn,
    op,
)
```

aid same result?

src/puyapy_mocks/_primatives.py
```diff
@@ -0,0 +1,386 @@
...
+        return Bytes(self.__match_value__ + bytes(other))
+
+    def __eq__(self, other: Bytes | bytes) -> bool:  # type: ignore[override]
+        return self.__match_value__ == bytes(other)
+
+    def __int__(self) -> int:
+        return int.from_bytes(self.__match_value__)
+
+    def __repr__(self) -> str:
+        return self.__repr__()
```
aorumbayev, line 370: @neilcampbell isn't this infinite recursion? I think this might need a str object instead

src/puyapy_mocks/op.py
```diff
@@ -0,0 +1,67 @@
...
+)
+from puyapy_mocks._reference import (
+    Account,
+)
+
+
+def btoi(a: Bytes | bytes, /) -> UInt64:
+    if isinstance(a, Bytes):
+        a = bytes(a)
+    return UInt64(int.from_bytes(a))
```
aorumbayev, line 14: byteorder is 'big' by default ?

src/puyapy_mocks/_reference.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __ne__(self, other: Account | str) -> bool:  # type: ignore[override]
+        return True
+
+    def __bool__(self) -> bool:
+        return True
+
+    @property
+    def balance(self) -> UInt64:
+        return UInt64(active_ctx().balances[self.address]["0"])
```
aorumbayev, line 22: 0 always assumes algo balance?

src/puyapy_mocks/_ctx_state.py
```diff
@@ -0,0 +1,31 @@
...
+    args: list[bytes] = attrs.field(factory=list)
+    logs: list[bytes] = attrs.field(factory=list)
+    balances: dict[str, dict[str, int]] = attrs.field(factory=dict)
+
+    def adjust_balance(self, address: str, asset_id: int, amount: int) -> None:
+        if address not in self.balances:
+            self.balances[address] = {}
+        if str(asset_id) not in self.balances[address]:
+            self.balances[address][str(asset_id)] = 0
+        self.balances[address][str(asset_id)] += amount
```
aorumbayev, line 15: wouldn't using asset_id as int instead of str when used as a key in a dict be a simpler option? Given that by default algod and indexer return the field as a 'number'

src/puyapy_mocks/op.py
```diff
@@ -0,0 +1,67 @@
...
+
+class AssetHoldingGet:
+    @staticmethod
+    def asset_balance(a: Account | UInt64 | int, b: UInt64 | int, /) -> tuple[UInt64, bool]:
+        if isinstance(a, Account):
+            address = a.address
+        elif isinstance(a, (int, UInt64)):
+            address = str(a)
+
+        return UInt64(active_ctx().get_balance(address, int(b))), False
```
aorumbayev, line 67: @neilcampbell what happens when a is of type UInt64? Wouldn't address be None in this case ?

src/puyapy_mocks/_primatives.py
neilcampbell: It sure is. That should have been `.__str__()`

src/puyapy_mocks/__init__.py
neilcampbell: Yep, it was to appease the ruff formatter from detecting that the import wasn't used and deleting it. I've since set noqa as the import order is important, so not really needed.

src/puyapy_mocks/op.py
neilcampbell: Yep, it's big endian by default.

src/puyapy_mocks/_reference.py
neilcampbell: Most of these implementations were to just make it work, so will need to be properly implemented when we build that outcome for real.

src/puyapy_mocks/_ctx_state.py
neilcampbell: Yep it would.

src/puyapy_mocks/op.py
neilcampbell: Yeah this code is very wrong, so we'll need to handle int/UInt64 address values properly.

neilcampbell:
Closing this, as it has served it's purpose.

PR #125, opened 2024-03-12
feat: add optimization to propagate constants found in Phi nodes resolving to the same constant value

daniel-makerx:


src/puya/ir/models.py
```diff
@@ -100,6 +100,8 @@ def _frozen_data(self) -> object:
 class Constant(Value, abc.ABC):
     """Base class for value constants - any value that is known at compile time"""
 
+    value: object
```
achidlow, line 103: shouldn't need this, these classes are value-equatable, don't need to peer inside them.

src/puya/ir/optimize/constant_propagation.py
```diff
@@ -42,9 +42,29 @@ def gather_constants(subroutine: models.Subroutine) -> dict[models.Register, mod
...
+            if phi_constant := _get_singular_phi_constant(phi, constants):
+                constants[phi.register] = phi_constant
+
     return constants
 
 
+def _get_singular_phi_constant(
+    phi: models.Phi, constants: dict[models.Register, models.Constant]
+) -> models.Constant | None:
+    constant: models.Constant | None = None
```
achidlow, line 56:
```python3
try:
    constant, = {constants[phi_arg.value] for phi_arg in phi.args}
except (KeyError, ValueError):
    return None
else:
    return constant
```
Should do the same thing and be a bit quicker?

daniel-makerx: Thanks

src/puya/ir/models.py
daniel-makerx: Yep

achidlow (approved):

PR #128, opened 2024-03-13
feat: allow using the throwaway ('_') var

daniel-makerx:


tristanmenzel (approved):

src/puya/awst_build/subroutine.py
```diff
@@ -1146,11 +1160,14 @@ def visit_bytes_expr(self, expr: mypy.nodes.BytesExpr) -> Literal:
 
     def visit_tuple_expr(self, mypy_expr: mypy.nodes.TupleExpr) -> ExpressionBuilder:
         items = [
-            require_expression_builder(
-                mypy_item.accept(self),
-                msg="Python literals (other than True/False) are not valid as tuple elements",
-            ).rvalue()
-            for mypy_item in mypy_expr.items
+            eb.lvalue() if self._is_lvalue_context else eb.rvalue()
```
achidlow, line 1163: pretty sure this will allow referencing a `_` variable if inside, say, a function call that occurs on the LHS of an assignment, or any other situation where an rvalue context is entered from within an lvalue one

PR #130, opened 2024-03-15
fix: don't inline control ops of Switch or GotoNth nodes if it would result in additional copies in destructured SSA

daniel-makerx:


src/puya/ir/optimize/control_op_simplification.py
```diff
@@ -183,6 +187,10 @@ def remove_target(parent: models.BasicBlock, to_remove: models.BasicBlock) -> No
             ) as fallthrough_terminator if (
                 # only if the default_block is empty
                 not (default_block.ops or default_block.phis)
+                # only if default_blocks targets don't have phi nodes,
+                # as this can result in excessive copies when ssa is destructured.
+                # re-evaluate this condition when parallel copy algorithm is improved
```
achidlow, line 192: it's not the parallel copy algorithm, it's the destructuring algorithm as a whole

achidlow (approved):

PR #136, opened 2024-03-18
Native UTF-8 String type

achidlow:
Also fixes an optimizer bug where differing behaviour of extract with & without immediates when length is zero was not accounted for


src/puya/awst/wtypes.py
```diff
@@ -545,7 +551,7 @@ def arc4_to_avm_equivalent_wtype(arc4_wtype: WType) -> WType:
         case ARC4DynamicArray(element_type=ARC4UIntN(n=8)):
             return bytes_wtype
     if arc4_wtype is arc4_string_wtype:
-        return bytes_wtype
+        return bytes_wtype  # TODO: this should be string_wtype now
```
daniel-makerx, line 554: Whats needed to do this TODO?

achidlow: @daniel-makerx will require either changing the base type of _ABIEncoded in the stubs etc, or paring that type down as discussed, hence leaving this for another PR

daniel-makerx (approved):

PR #140, opened 2024-03-19
App mode validation

tristanmenzel:
- Validate ops used in a contract/logic sig are available for those respective modes
 - Refactor avm version validation to match the same pattern


src/puya/ir/types_.py
```diff
@@ -58,8 +58,7 @@ def wtype_to_avm_type(
         case wtypes.void_wtype:
             raise InternalError("Can't translate void WType to AVMType", source_location)
         case _:
-            # TODO: make this a CodeError
-            raise InternalError(
+            raise CodeError(
```
achidlow, line 61: resolving this TODO requires updating the call site as well ;)

achidlow (approved):

src/puya/ir/types_.py
achidlow: fixed

PR #141, opened 2024-03-21
Simpler default construction of certain types, `Account` and `arc4.Address` constructor parity, and improve constant folding / intrinsic simplification

achidlow:
feat: Allow empty constructors to default to "zero" values for UInt64, BigUInt, Account, Asset, Application, arc4.Address, and arc4 numeric types. Also, generally establish feature parity between Account and arc4.Address constructors

feat: simplify more conditions they're the target of a, `assert`, `select`, or `bz`/`bnz` op

feat: allow `==`/`!=` comparison between `arc4.Address` and `Account`

feat: allow bytes optimizations to handle `addr` constants, and also `global ZeroAddress` ops

feat: allow `bytes` literal with `BytesBacked.from_bytes`

fix: add some missing positional-only indicators to stubs, and relax typing.LiteralString to str, it's not really applicable for our use case

other minor todos, and small tests added


examples/tictactoe/tictactoe.py
```diff
@@ -51,9 +51,7 @@ def play(self, move: Move) -> None:
             assert Txn.sender == self.host, "It is the host's turn"
             player = arc4.UInt8(HOST)
         else:
-            assert Txn.sender == self.challenger.get(
-                Global.zero_address
-            ), "It is the challenger's turn"
+            assert Txn.sender == self.challenger.get(Account()), "It is the challenger's turn"
```
robdmoore, line 54: This reads confusingly, like what is `Account` in this case. I can't look at this an intuitively know. Why is anything needed in `get` anyway?

examples/auction/contract.py
```diff
@@ -17,9 +18,8 @@ def __init__(self) -> None:
         self.auction_end = UInt64(0)
         self.previous_bid = UInt64(0)
         self.asa_amount = UInt64(0)
-        self.asa = Asset(0)
-        # Use zero address rather than an empty string for Account type safety
-        self.previous_bidder = Global.zero_address
+        self.asa = Asset()
+        self.previous_bidder = Account()
```
tristanmenzel, line 22: I'm not sold on this change. Could we do `Account.zero_address` ?

examples/tictactoe/tictactoe.py
tristanmenzel: The parameter is the default value if the state var is not set, but I agree that `Account()` is confusing

daniel-makerx: Could be simplified by making `self.challenger` use Global.zero_address as the unset/initial value?

src/puya/ir/optimize/intrinsic_simplification.py
```diff
@@ -336,16 +382,31 @@ def _choose_encoding(a: AVMBytesEncoding, b: AVMBytesEncoding) -> AVMBytesEncodi
     return AVMBytesEncoding.base16
 
 
+def _decode_address(address: str) -> bytes:
+    """check if address is a valid address with checksum"""
```
daniel-makerx, line 386: is this docstring incorrect? or should this function validate the length and hash?

src/puyapy-stubs/_reference.pyi
```diff
@@ -254,8 +256,8 @@ class Asset:
 class Application:
     """An Application on the Algorand network."""
 
-    def __init__(self, application_id: UInt64 | int, /):
-        """Initialized with the id of an application"""
+    def __init__(self, application_id: UInt64 | int = 0, /):
+        """Initialized with the id of an application. Default to zero (an invalid ID)."""
```
daniel-makerx, line 260:
```suggestion
        """Initialized with the id of an application. Defaults to zero (an invalid ID)."""
```

src/puyapy-stubs/arc4.pyi
```diff
@@ -350,7 +350,14 @@ class DynamicArray(_ABIEncoded, typing.Generic[_TArrayItem], Reversible[_TArrayI
 class Address(StaticArray[Byte, typing.Literal[32]]):
     """An alias for an array containing 32 bytes representing an Algorand address"""
 
-    def __init__(self, account_or_bytes: puyapy.Bytes | puyapy.Account | bytes, /): ...
+    def __init__(self, value: puyapy.Account | str | puyapy.Bytes = ..., /):
```
daniel-makerx, line 353: No `bytes` literal anymore?

tests/test_execution.py
```diff
@@ -1276,3 +1276,43 @@ def test_inheritance_direct_method_invocation(harness: _TestHarness) -> None:
         "ChildContract.method called",
         "GrandParentContract.method called",
     ]
+
+
+def test_account_from_bytes_validation(harness: _TestHarness) -> None:
+    def test() -> None:
```
daniel-makerx, line 1282: I thought we stopped doing tests from closures?

examples/auction/contract.py
robdmoore: I think `Account()` as zero address makes sense, but I agree having the more explicit `Account.zero_address` as an option would be more readable as an option?

src/puya/ir/optimize/intrinsic_simplification.py
achidlow: copy-pasta error

src/puyapy-stubs/arc4.pyi
achidlow: It didn't make sense here, the address `str` is much more usable.
I'll add `bytes` literals to `from_bytes`.

tests/test_execution.py
achidlow: Why's that?

daniel-makerx: You removed them previously (I forget why), they also are a bit slower since they aren't part of the AWST caching

examples/auction/contract.py
achidlow: There is a convention in Python where the default/empty constructor yields an empty/"Falsey" value:
- `int() == 0`
- `bool() == False`
- `str() == ""` 
and so on.
As part of this PR, I've made this change to our types, which actually reads a lot nicer where you don't have to explicitly put zero like `Asset(0)` to get an "empty"/`False` `Asset`.
It would be strange if `Account` didn't follow this convention and everything else did.

(side note: I had assumed UInt64 and BigUInt already had these default constructor behaviours for us, just noticed they haven't, adding that now)

examples/tictactoe/tictactoe.py
achidlow: so `.get()` in Python mappings would return `None` if the key isn't present, that's the same idea here with `GlobalState` proxies. I'll add `default=` in this example, but using `GlobalState` instead here vs initialising to zero like you've suggested @daniel-makerx was what the contract initially did, the way it is not generates more efficient code so is useful to point out in an example.

To someone fluent in Python `.get()` and the default arg makes total sense and this reads well to me at least, although if AVM had enough support that we could utilise `None` it'd be even better.

tests/test_execution.py
achidlow: I don't recall doing that 🤔

Anyway, these ones are here to avoid too much cruft in `test_cases`, since they're basically one liners.

examples/auction/contract.py
achidlow: Another side note @tristanmenzel, with this consistency, we can make a sensible "default empty" constructor for arc4 static arrays too. Will be much easier to do with a bit of EB refactoring too that I'm holding off on until post 1.0

daniel-makerx (approved):

PR #149, opened 2024-03-25
ARC4 semantic issues workarounds

achidlow:


tests/test_execution.py
```diff
@@ -1333,7 +1334,50 @@ def clear_state_program(self) -> bool:
                 return True
 
     # expected error message here is just an example, update test once a better one is available
-    import puya.errors
-
     with pytest.raises(puya.errors.CodeError, match="Invalid use of a type union"):
         harness.deploy_from_closure(test)
+
+
+def test_mutable_ref_param_reassignment_fails(harness: _TestHarness) -> None:
```
daniel-makerx, line 1341: These would be better as part of the expected_errors.test suite?

tests/test_execution.py
```diff
@@ -1333,7 +1334,50 @@ def clear_state_program(self) -> bool:
...
+    def test() -> None:
+        from puyapy import arc4
+
+        class Baddie(arc4.ARC4Contract):
+            @arc4.abimethod
+            def okay(self, arr: arc4.DynamicBytes) -> None:
+                arr = arc4.DynamicBytes.from_bytes(arr.bytes)
```
tristanmenzel, line 1348: Probably worth adding a comment explaining _why_ this assignment is ok

tristanmenzel (approved):

tests/test_execution.py
achidlow: Because I thought (wrongly) that those tests didn't lower to IR, where this error is generated.

daniel-makerx: I've moved them as I suggested

PR #157, opened 2024-03-27
Version and stub discover fixes

daniel-makerx:


src/puya/compile.py
```diff
@@ -172,50 +170,61 @@ def module_irs_to_teal(
...
+def _get_python_executable() -> str | None:
+    prefix = _get_prefix()
+    if not prefix:
+        logger.warning("Could not determine python prefix or algopy version")
+        return None
+    logger.info(f"Found python prefix: {prefix}")
+    venv_paths = sysconfig.get_paths(vars={"base": prefix})
+
+    # use sys.executable as cross-platform way of correctly using python.exe or python
+    python_exe = Path(venv_paths["scripts"]) / Path(sys.executable).name
```
achidlow, line 182: There's no guarantee that the `sys.executable` name will match that of the target environment

src/puya/compile.py
```diff
@@ -172,50 +170,61 @@ def module_irs_to_teal(
...
-            )
-        else:
-            logger.debug(f"Using python site-packages: {site_packages}")
-    _check_algopy_version(site_packages)
-    return str(python_exe) if python_exe else None
+    else:
+        logger.debug(f"Using python site-packages: {site_packages}")
+        _check_algopy_version(site_packages)
 
+    return str(python_exe)
```
achidlow, line 204: this could be `str(None)`

achidlow (approved):

src/puya/compile.py
daniel-makerx: oops

PR #163, opened 2024-03-28
docs: add algorand-python instruction to quick start

neilcampbell:


neilcampbell:
@daniel-makerx Should we merge as a `chore` or `docs`. I believe we'll want to avoid a release.

robdmoore:
Superceeded by #164

PR #174, opened 2024-04-04
feat: Add support for bool types in state proxies

### LINKED ISSUE #175: Support bool type in GlobalState and LocalState proxies
The generic type constraint on the stubs does not allow for a boolean state value

```python
_TState = typing.TypeVar("_TState", bound=UInt64 | Bytes | BytesBacked)
```

We should modify the stubs to allow `bool` and make any other required changes to ensure bool state proxies work

tristanmenzel:
fixes #175


robdmoore:
Looks good! Needs docs though

stubs/algopy-stubs/_state.pyi
```diff
@@ -2,7 +2,7 @@ import typing
 
 from algopy import Account, Bytes, BytesBacked, UInt64
 
-_TState = typing.TypeVar("_TState", bound=UInt64 | Bytes | BytesBacked)
+_TState = typing.TypeVar("_TState", bound=UInt64 | Bytes | BytesBacked | bool)
```
daniel-makerx, line 5: This makes me release we might need a `UInt64Backed` for things like Asset and Application, or at least include them in this TypeVar

As currently this works

`self.some_asset = Asset(....)`

But this won't
```
self.some_asset = GlobalState(Asset)
self.some_asset.value = Asset(...)
```

achidlow: I think we either expand the type var bounds or just drop the bounds.

Not sure a `UInt64Backed` makes sense in the same way that `BytesBacked` does.

achidlow (approved):

PR #176, opened 2024-04-04
fix: Ensure arc4 router generation works with minimal contracts that don't have any explicit abimethods

### LINKED ISSUE #173: Support ARC4 contract with no explicit methods
The following minimal arc4 contract should compile but currently does not as we only generate the router when there are `@abimethod` decorated functions. However even when there are no explicit methods, we should still generate an implicit create and clear_state implementation.

```python
from algopy import ARC4Contract, UInt64

class MinimumARC4(ARC4Contract):
    def __init__(self) -> None:
        self.gvalue = UInt64(4)
```

The reported error is `critical: contract is non abstract but doesn't have approval and clear programs in hierarchy`. 

The bug stems from [this](https://github.com/algorandfoundation/puya/blob/1a29ad635228df220a60bd6a55fd30998dbca897/src/puya/ir/main.py#L423) line

tristanmenzel:
fixes #173


src/puya/ir/arc4_router.py
```diff
@@ -698,12 +701,7 @@ def create_abi_router(
         source_location=router_location,
         args=[],
         return_type=wtypes.bool_wtype,
-        body=create_block(
-            router_location,
-            "abi_bare_routing",
-            *router,
-            reject(contract.source_location),
-        ),
+        body=create_block(router_location, "abi_bare_routing", *router),
```
tristanmenzel, line 704: I removed the 'fall through' `reject` here in favour of explicitly terminating the blocks created by `route_abi_methods`

test_cases/state_totals/out/Contract.approval.teal
```diff
@@ -7,17 +7,8 @@ test_cases.state_totals.contract.Contract.approval_program:
     //     state_totals=StateTotals(local_bytes=1, global_uints=GLOBAL_UINTS),
     // ):
     txn NumAppArgs
-    bz main_bare_routing@4
-    txna ApplicationArgs 0
-    match
```
tristanmenzel, line 12: This appears to be the only change to an optimised teal file, and it actually fixes a bug we'd missed where an empty `match` is created

src/puya/ir/main.py
```diff
@@ -420,7 +420,13 @@ def fold_state_and_special_methods(
     if not (c.init and c.init.body.body):
         result.init = None
     arc4_method_refs = dict(filter(None, maybe_arc4_method_refs.values()))
-    if arc4_method_refs:
+    if arc4_method_refs and not contract.is_arc4:
+        raise CodeError(
```
daniel-makerx, line 424: This should be checked earlier during AWST build, and an error is emitted then if and arc4 decorator is used with a non arc4 class, so this might make more sense as an InternalError. To be sure should also add a test for this case to tests/test_expected_output/arc4.test

test_cases/state_totals/out/Contract.approval.teal
achidlow: Haha, looks like it still runs, but yeah it's effectively a no-op that costs 2 ops 😬

achidlow (approved):

PR #188, opened 2024-04-08
fix: handle arc4.Address.native

daniel-makerx:


test_cases/arc4_types/address.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Arc4AddressContract(Contract):
+    def approval_program(self) -> bool:
+        address = arc4.Address(Txn.sender)
+        assert address == Txn.sender
+        assert address.length == 32
+        assert address.native == Txn.sender
+
+        zero_address = arc4.Address(Global.zero_address.bytes)
```
tristanmenzel, line 13: I think we can do `arc4.Address()` now right?

tristanmenzel: (to get a zero address)

daniel-makerx: Yeah, but in this case I wanted to cover the constructor case that takes a bytes, `Global.zero_address` was just a convenient source of another address

tristanmenzel: ah ok, I just use `Txn.sender.bytes` :)

tristanmenzel (approved):

PR #192, opened 2024-04-12
Improve test coverage + minor fixes/enhancements

### LINKED ISSUE #191: Generated `algopy.op` stubs for variadic functions sometimes have incorrect argument types
Bug in the generation script, appears to only affect `setbit` variants:

```python3
def setbit_bytes(a: Bytes | bytes | UInt64 | int, b: UInt64 | int, c: UInt64 | int, /) -> Bytes:
```
Above should not include `UInt64 | int` as valid types for `a`.

```python3
def setbit_uint64(a: Bytes | bytes | UInt64 | int, b: UInt64 | int, c: UInt64 | int, /) -> UInt64:
```
Above should not include `Bytes | bytes` as valid types for `a`.

achidlow:
Fixes #191


src/puya/awst/nodes.py
```diff
@@ -786,23 +787,24 @@ class TupleItemExpression(Expression):
 
     Note: this is its own item (vs IndexExpression) for two reasons:
     1. It's not a valid lvalue (tuples are immutable)
-    2. The index must always be a literal
+    2. The index must always be a literal, and can be negative
     """
 
     base: Expression
-    index: int
+    index: int = attrs.field(validator=attrs.validators.ge(0))
```
tristanmenzel, line 794: Doesn't this validator go against point 2. above?

tristanmenzel: I see the eb is converting it to a positive index, so perhaps it's point 2. that needs rewording

tristanmenzel (approved):

src/puya/awst/nodes.py
achidlow: ah, this can actually be removed, wasn't required in the end

PR #203, opened 2024-04-22
chore: setting up base sub project structure for unit testing package

aorumbayev:
## Proposed Changes

- Setting up base structure for an algopy_testing package. Assumes that existing structure on the repo is to be further re grouped to contain 3 independent poetry projects at the root of the monorepo (stubs and compiler source folder to be slightly renamed + regrouped). 

## ADR Recap

The following establishes a starting point for implementation that is described in several steps.

### 1. Base structure

The implementation is to be placed in a sub project making the repo root a monorepo. The current package name is `algorand-python-testing` and it is to be further refined as needed, one option is to make it available for installation as an extra for `algorand-pythoh` package (allowing `algorand-python[testing]` extras). The actual base structure can be observed in the pull request. The gist of the approach is to implement mocks/logic matching avm behaviour (such as crypto primitives, math operations and etc). 

By ensuring that algorand-python-testing ships a module named algopy (identical to exposed algopy module that algorand-python supplies) we can rely on stub substitution that will inject implementation code allowing users to execute their python smart contracts in test contexts. 

### 2. Main outcome requirements

To satisfy main requirements - stubs in algorand-python will be implemented - primarily focusing on allowing users to mock anything around application, subroutines, logicsigs, while primitive types won't require mocking (as they will evaluate matching python code) as well as the crypto primitives (for which, a matching behaviour can be achieved using `nacl` and `coincurve` packages, needed for Ecdsa and Ed25519, as outlined in https://pyteal.readthedocs.io/en/stable/crypto.html)

Additionally a set of helper abstractions are to be implemented to simulate the necessary state/ledge management related behaviour

### 3. Property based testing 

Outlined approach works and fits nicely with https://hypothesis.readthedocs.io/en/latest/, framework, the results of this spike confirmed that integrating hypothesis is a simple and straightforward process that can involve either defining custom `strategies` or having users leverage https://hypothesis.readthedocs.io/en/latest/data.html#composite-strategies strategies to define custom behaviour for native AVM types. Initial implementation will include a set of documentation code snippets that users can reuse. If need/feedback arises from the community to have custom hypothesis strategies be shipped as part of the testing package, that specific logic can be distributed via `extras`.


algopy_testing/__init__.py
daniel-makerx, line 1: I don't think we need this one now? All source should be under src/

algopy_testing/pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+[tool.poetry.dependencies]
+python = "^3.12"
+
+[tool.poetry.group.testing.dependencies]
+pytest = "^7.4.0"
+pytest-xdist = {extras = ["psutil"], version = "^3.3.1"}
+py-algorand-sdk = "^2.4.0"
+algokit-utils = "^2.2.1"
+pytest-cov = "^4.1.0"
+prettytable = "^3.9.0"
```
daniel-makerx, line 25: at this stage I don't think we will need prettytable, it was used for one very specific thing in the puya tests

algopy_testing/src/algopy/__init__.py
```diff
@@ -0,0 +1,14 @@
+# ruff: noqa
+
+# TODO: uncomment as new features are implemented
```
daniel-makerx, line 3: Rather than a bunch of TODOs I was thinking of having a script that checks the stubs against the testing implementations and outlines any stubs that aren't covered.

pyproject.toml
```diff
@@ -115,6 +115,7 @@ exclude = [
     "src/puya/_vendor",
     "stubs/algopy\\.py",
 ]
+files = ['stubs/*', 'src/*']
```
daniel-makerx, line 118: I think it will better for now to just add algopy_testing as another exclusion rather than modifying files

.pre-commit-config.yaml
daniel-makerx, line 1: can we do this change without reformatting the file?

aorumbayev: @daniel-makerx tweaked editorconfig to match the formatting that was present on the yaml

daniel-makerx (approved):

PR #207, opened 2024-04-27
feat: UInt64 stub implementation

aorumbayev:
## Proposed Changes

The following PR focuses on setting up initial implementation for UInt64 AVM primitive type. After further discussions with Dan, we decided to put more focus on getting solid coverage for this particular abstraction as it serves as a base for a large amount of consequent abstractions that we are to cover.

- Implement the UInt64 class, which represents a 64-bit unsigned integer, one of the primary data types on the Algorand Virtual Machine (AVM).
- Add error classes for handling various exceptions related to UInt64 operations, such as overflow, underflow, and invalid values.
- Provide comprehensive test cases for the UInt64 class, covering all arithmetic operations, bitwise operations, and edge cases using hypothesis framework.
- Update the README file with instructions for checking the coverage of the algopy stubs and regenerating test artifacts.
- Add a script to check the coverage of the algopy stubs against the 
- Propagate overlapping settings from main pyproject to algopy_testing in preparation for further root project restructure. Mainly ruff, vscode settings for ruff auto formatting, mypy settings. 
- Addressing all mypy and ruff warning and suggestions (note that you need algopy_testing folder to be opened in separate workspace as its a standalone poetry project)

## Further notes

- I meant to have this as a separate draft pr but didn't get enough time. The starter algokit project used during spike item for crypto primitives is temporarily stored in this private repo, can be used for reference while im off on vacation. Once i'm back ill transfer the useful snippets from there to algopy_testing and remove the private repo. https://github.com/algorandfoundation/puya-crypto-primitives-spike/blob/master/tests/crypto_primitives_test.py (see integration tests confirming matching behaviour with various avm primitivies)


algopy_testing/tests/utils.py
```diff
@@ -0,0 +1,6 @@
+def prioritize_local_algopy_over_stubs() -> None:
```
aorumbayev, line 1: @daniel-makerx curious to learn if there are better ways to achieve that

achidlow: This shouldn't be required? an `algopy/__init__.py` will take precedence over an `algopy.py`, which is what the stubs provides to keep PyLance happy.

algopy_testing/tests/primitives/test_uint64.py
```diff
@@ -0,0 +1,267 @@
...
+        UInt64(value)
+
+
+@given(
+    st.integers(min_value=0, max_value=MAX_UINT64), st.integers(min_value=0, max_value=MAX_UINT64)
+)
+@settings(max_examples=EXAMPLES_PER_AVM_TEST)
+def test_uint64_addition(primitive_ops_client: PrimitiveOpsContractClient, a: int, b: int) -> None:
+    """Test the addition operation for UInt64 instances."""
+    if a + b > MAX_UINT64:
```
achidlow, line 77: The way hypothesis works, this won't be guaranteed to be hit just because the range includes them.

I'm not sure hypothesis provides much advantage here, but if  it is used, we'll need to add specific examples for boundary conditions (hypothesis allows this with `@example`)

algopy_testing/tests/artifacts/PrimitiveOps/contract.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class PrimitiveOpsContract(ARC4Contract):
+    @arc4.abimethod()
+    def verify_uint64_init(self, raw_value: Bytes) -> UInt64:
+        return op.btoi(raw_value)
+
+    @arc4.abimethod()
+    def verify_uint64_add(self, a: UInt64, b: UInt64) -> UInt64:
+        return a + b
```
achidlow, line 11: if we split these into 
```python
result = a <op> b
return result
```
We can verify the failures occur at the right spot, not just at the end of the function

algopy_testing/src/algopy/primitives/uint64.py
```diff
@@ -0,0 +1,735 @@
...
+
+        Raises:
+            ValueError: If the shift amount is negative.
+
+        Returns:
+            UInt64: The result of the left shift operation.
+        """
+        shift = self._get_value(other)
+        if shift < 0:
+            raise UInt64BitShiftInvalidValueError("Shift amount must be non-negative")
```
achidlow, line 496: this doesn't need a specific error, just construct a UInt64 from the right value, which is underflow.
negative constants won't compile anyway, and we don't need to guarantee any semantic compatibility if the code doesn't compile

algopy_testing/src/algopy/primitives/uint64.py
```diff
@@ -0,0 +1,735 @@
...
+
+        Returns:
+            UInt64: The result of the subtraction.
+        """
+        result = other - self.value
+        if result < 0:
+            raise UInt64UnderflowError(result)
+        return UInt64(result)
+
+    def __isub__(self, other: int | UInt64) -> Self:
```
achidlow, line 239: the default implementation is fine, we don't need to implement this at all (or any other `__i..__` method), since we consider UInt64 as immutable

algopy_testing/src/algopy/primitives/uint64.py
```diff
@@ -0,0 +1,735 @@
...
+        declared at the module level.
+
+        Args:
+            value (int, optional): The initial value of the UInt64. Defaults to 0.
+
+        Raises:
+            UInt64InvalidValueError: If the provided value is not an integer.
+            UInt64UnderflowError: If the provided value is negative.
+            UInt64OverflowError: If the provided value exceeds the maximum value for a UInt64.
+        """
```
achidlow, line 35: do we need all these docstrings for what amounts to a mock class?

algopy_testing/src/algopy/primitives/uint64.py
```diff
@@ -0,0 +1,735 @@
...
+            Defaults to None.
+
+        Raises:
+            UInt64OverflowError: If the power result exceeds the maximum value for a UInt64.
+
+        Returns:
+            UInt64: The result of the power operation.
+        """
+        exp = self._get_value(other)
+        if modulo is not None:
```
achidlow, line 433: we don't support the `pow` function yet

algopy_testing/tests/utils.py
daniel-makerx: This turns out to be rather complex when using "editable" dependencies as the `algopy__init__.py` taking precedence over `algopy.py` only works if the both the `algorand-python` and `algorand-python-testing` packages are colocated in the same venv/path

Solution for now was to make the stubs a dependency only when packaging the testing package but not during develpoment

algopy_testing/tests/primitives/test_uint64.py
daniel-makerx: For now have just used regular pytest parameters, have left hypothesis as a dev dependancy for usages in more complex scenarios

algopy_testing/src/algopy/primitives/uint64.py
daniel-makerx: Have opted for using builtin python errors for now, we can always move towards more specific types if required

algopy_testing/tests/artifacts/PrimitiveOps/contract.py
daniel-makerx: Have made this change, will leave the work of verifying failures occur at the correct position for later (once subroutines and/or contracts are testable)

algopy_testing/src/algopy/primitives/uint64.py
daniel-makerx: I verified that when the implementation and stubs are installed in a venv together that both PyCharm and VSCode will show the stub documentation over the implementation, so docstrings aren't required.

algopy_testing/src/algopy/primitives/uint64.py
daniel-makerx: For now am returning `NotImplemented` if a modulo is specified

algopy_testing/src/algopy/primitives/uint64.py
daniel-makerx: Removed the assignment operator implementations, and added tests verifying the default behaviour

PR #211, opened 2024-05-09
refactor: Introduce IntersectionSliceExpression awst node which performs internal bounds checking

tristanmenzel:
## Proposed Changes

- Introduce a new node IntersectionSliceExpression  which represents a slice expression where bounds are checked at the IR level.


achidlow (approved):
Nice

PR #212, opened 2024-05-10
feat: Partial Bytes stub implementation

boblat:
- bitwise operators are going to be implemented in a follow-up PR
- tests where hitting the 4096 bytes length limit have been commented out for now as AVM is throwing `"application args total length too long, max len 2048 bytes" error`


algopy_testing/src/algopy/primitives/bytes.py
```diff
@@ -0,0 +1,167 @@
...
+    def __str__(self) -> str:
+        return str(self.value)
+
+    def __bool__(self) -> bool:
+        return bool(self.value)
+
+    def __add__(self, other: Bytes | bytes) -> Bytes:
+        """Concatenate Bytes with another Bytes or bytes literal
+        e.g. `Bytes(b"Hello ") + b"World"`."""
+        return (
```
daniel-makerx, line 47: Despite the typing hints, I think we need to be conservative here and, attempt to convert other to bytes first, so the appropriate error can be thrown if the wrong type was passed. 

e.g.
```
result = self.value + as_bytes(other)
return _checked_result(result, "+")
```

algopy_testing/src/algopy/primitives/bytes.py
```diff
@@ -0,0 +1,167 @@
...
+        # my_bytes[0:1] => b'j' whereas my_bytes[0] => 106
+        return (
+            Bytes(self.value[index])
+            if isinstance(index, slice)
+            else Bytes(self.value[slice(index, index + 1)])
+        )
+
+    def __eq__(self, other: object) -> bool:
+        maybe_bytes = _as_maybe_bytes(other)
+        # returning NotImplemented here will allow BigUInt to handle larger comparisons
```
daniel-makerx, line 83: This comment doesn't apply for Bytes since we don't do any up-casting, instead should use `_as_bytes(other)` as it would be a compile error to compare to other types

algopy_testing/src/algopy/primitives/bytes.py
```diff
@@ -0,0 +1,167 @@
...
+        case _:
+            return None
+
+
+def _checked_result(result: bytes, op: str) -> Bytes:
+    """Ensures `result` is a valid Bytes value
+
+    Raises:
+        ArithmeticError: If `result` of `op` is out of bounds"""
+    if len(result) < 0:
```
daniel-makerx, line 157: We can remove this check, as `len` should always be non-negative

algopy_testing/tests/primitives/test_bytes.py
```diff
@@ -0,0 +1,42 @@
...
+
+@pytest.mark.parametrize(
+    ("a", "b"),
+    [
+        (b"", b""),
+        (b"1", b""),
+        (b"", b"1"),
+        (b"1", b"1"),
+        # throws "application args total length too long, max len 2048 bytes" error
+        # (b"", b"0" * MAX_BYTES_SIZE),
```
daniel-makerx, line 29: As mentioned we should be able to use `op.bzero` to pad arguments so we can reach sizes > 2048, the python equivalent to `op.bzero(some_length)` would be 
`b"\x00" * some_length`

algopy_testing/src/algopy/primitives/bytes.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def _as_maybe_bytes(value: object) -> bytes | None:
+    """Returns bytes value if `value` is a bytes or Bytes, otherwise None"""
+    match value:
+        case bytes(bytes_value):
+            return _as_bytes(bytes_value)
+        case Bytes(value=bytes_value):
+            return bytes_value
+        case n if (isinstance(n, Sequence) and all(isinstance(i, int) for i in n)):
```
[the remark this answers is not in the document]
boblat: in the tests where we assert equality, e.g. `assert avm_result == Bytes(a) + Bytes(b)`, it calls `__eq__` method where `other: object` parameter is given a list of int values. Without this case, we get a `TypeError`.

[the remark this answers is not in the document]
boblat: It has been fixed in [this commit](https://github.com/algorandfoundation/puya/pull/212/commits/e93ecdf45b6a937a5be7eb47ad307e5db8bcd35f) as discussed.

algopy_testing/tests/primitives/test_bytes.py
boblat: It has been implemented in [this commit](https://github.com/algorandfoundation/puya/pull/212/commits/84c9cf38e63f65041ad12d18afa5ab5edf61fc06) as discussed.

PR #216, opened 2024-05-14
feat: bitwise operators and tests for Bytes stub implementation

boblat:
## Proposed Changes

- added bitwise operations for Bytes
- added additional tests for Bytes stub implementation


algopy_testing/tests/artifacts/PrimitiveOps/contract.py
```diff
@@ -138,6 +138,8 @@ def verify_bytes_xor(self, a: Bytes, b: Bytes) -> Bytes:
         return result
 
     @arc4.abimethod()
-    def verify_bytes_not(self, a: Bytes) -> Bytes:
+    def verify_bytes_not(self, a: Bytes, pad_size: UInt64) -> Bytes:
+        a = op.bzero(pad_size) + a
```
aorumbayev, line 142: I assume thats what allows skipping the arg size limit on input parameters?

algopy_testing/tests/primitives/test_bytes.py
```diff
@@ -79,5 +80,180 @@ def test_bytes_addition_overflow(
...
+        assert x1 == _int_to_bytes(x2)
+
+
+@pytest.mark.parametrize(
+    "index",
+    [
+        0,
+        slice(0, 2),
+        11,
+        slice(11, 1),
```
aorumbayev, line 247: wouldnt slice(11,1) always yield in empty slice? What was the edge case with this test input?

algopy_testing/src/algopy/primitives/bytes.py
```diff
@@ -80,6 +75,24 @@ def __getitem__(
...
+        return self._operate_bitwise(other, "and_")
+
+    def __rand__(self, other: bytes) -> Bytes:
+        return self & other
+
+    def __or__(self, other: bytes | Bytes) -> Bytes:
+        return self._operate_bitwise(other, "or_")
+
+    def __ror__(self, other: bytes) -> Bytes:
+        return self | other
```
aorumbayev, line 88: @daniel-makerx should there be a check on whether other is indeed of type bytes?

algopy_testing/src/algopy/primitives/bytes.py
```diff
@@ -1,11 +1,14 @@
...
 import base64
+import operator
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
+from itertools import zip_longest
+
 from algopy._constants import MAX_BYTES_SIZE, MAX_UINT64
```
aorumbayev, line 12: @daniel-makerx should we rename it to constants from _constants if its meant to be accessed from sibling namespaces like tests? I could also imagine this being useful in testing scenarios, constants like max byte size and etc would be handy to be accessible via the package instead of having to redefine them

algopy_testing/tests/artifacts/PrimitiveOps/contract.py
boblat: Yes. That is the approach we have taken for getting around the arg size limit on input parameters.

algopy_testing/tests/primitives/test_bytes.py
boblat: Thanks for finding that. I meant to `slice(11, 12)`.

algopy_testing/src/algopy/primitives/bytes.py
daniel-makerx: Can rename this to constants, however we will need a different approach to expose to consumers as the stubs definitions won't contain these declarations.

I'm thinking we will need something like an `algopy_testing` module that is provided in the wheel alongside these stubs implementations to provide additional functionality that isn't declared in the stubs. Such as utility functions, the testing context and things like these constants.

algopy_testing/src/algopy/primitives/bytes.py
daniel-makerx: This should be fine, as these bitwise operations are all commutative, so we can just forward the implementation on to the reflected operations which do the type check

algopy_testing/tests/primitives/test_bytes.py
```diff
@@ -79,5 +80,181 @@ def test_bytes_addition_overflow(
...
+    base64_str = "RkY="
+    assert Bytes.from_base64(base64_str) == base64.b64decode(base64_str)
+
+    hex_str = "FF"
+    assert Bytes.from_hex(hex_str) == base64.b16decode(hex_str)
+
+
+@pytest.mark.parametrize("value", [b"hello, world", b"", b"0110", b"0" * MAX_BYTES_SIZE])
+def test_bytes_len(value: bytes) -> None:
+    assert len(Bytes(value)) == len(value)
```
daniel-makerx, line 226: Should validate Bytes.length here too

algopy_testing/tests/primitives/test_bytes.py
```diff
@@ -79,5 +80,181 @@ def test_bytes_addition_overflow(
...
+
+
+@pytest.mark.parametrize("value", [b"hello, world", b"", b"0110", b"0" * MAX_BYTES_SIZE])
+def test_bytes_iter(value: bytes) -> None:
+    for x1, x2 in zip(iter(Bytes(value)), iter(value), strict=False):
```
daniel-makerx, line 231: I think these should be strict=True?

algopy_testing/src/algopy/primitives/bytes.py
boblat: Cool. I will remove the reference to `algo-y._constants` and re-implement them in tests just like in uInt64 test for now.

daniel-makerx (approved):

PR #217, opened 2024-05-14
feat: crypto ops implementation

aorumbayev:
## Proposed Changes

- Adds initial implementation for sha256, sha3_256, sha512_256, ed25519verify_bare and ed25519verify
- Making a few global fixtures more generic (for reuse with different typed clients)

## Open questions 

The ed25519verify_bare and ed25519verify are almost identical except that data in ed25519verify is concatenation of 'ProgData' + hash of current app + data passed by user, while bare takes the data passed by user and verifies it directly. @achidlow @daniel-makerx currently im forwarding the calls to verify to rely on verify_bare without appending the hash since an app isn't deployed or has access to avm context at that stage:

```
def ed25519verify(a: Bytes | bytes, b: Bytes | bytes, c: Bytes | bytes, /) -> bool:
    return ed25519verify_bare(a, b, c)
```

What if additionally we add an optional implementation specific argument for letting user to provide execution context specific data (or **mock** data given that it will be mostly in the unit testing contexts), making signature something like:
```
def ed25519verify(a: Bytes | bytes, b: Bytes | bytes, c: Bytes | bytes, /, program_hash: bytes = b'dummy_hash') -> bool:
    a = b'ProgData' + program_hash + a # if user does not explicitly set the hash it is still concatenated but with dummy content
    ... rest of the method
```

## TODO

- [x] - Clarify open question on ed25519verify vs ed25519verify_bare
- [x] - Implement EcdsaVerify
- [x] - Implement EcdsaDecompress
- [x] - Implement EcdsaRecover
- [ ] - Refine tests for edge cases


algopy_testing/src/algopy/op.py
```diff
@@ -0,0 +1,68 @@
...
+
+import nacl.exceptions
+import nacl.signing
+from Cryptodome.Hash import keccak
+
+from algopy import Bytes
+
+
+def sha256(a: Bytes | bytes, /) -> Bytes:
+    if not isinstance(a, bytes | Bytes):
```
daniel-makerx, line 11: We can move `algopy.primitives._utils` to `algopy.utils` and then use `as_bytes` here to get a `bytes` that has been checked for type and length

daniel-makerx:
>What if additionally we add an optional implementation specific argument for letting user to provide execution context specific data (or mock data given that it will be mostly in the unit testing contexts), making signature something like:

I don't think we can use this approach, as this means the smart contract a user writes would need to include this additional argument which wouldn't be accepted by the puya compiler and would be introducing test specific code into the smart contract.

Instead I think this sort of information should be part of the context information of a test, that our implementation can access and include when doing this op.

algopy_testing/src/algopy/op.py
aorumbayev: Done

aorumbayev:
@daniel-makerx see latest commit for initial implementation of generic context manager. Adding initial coverage for ecdsa as well however will do some extra polishing tomorrow, let me know if you got any comments in the meantime 👍

algopy_testing/src/algopy/op.py
```diff
@@ -0,0 +1,159 @@
...
+from algopy.utils import as_bytes
+
+
+class ECDSA(Enum):
+    Secp256k1 = 0
+    Secp256r1 = 1
+
+
+def sha256(a: Bytes | bytes, /) -> Bytes:
+    input_value = as_bytes(a, max_size=MAX_BYTES_SIZE)
```
daniel-makerx, line 28: we can make `max_size` default to MAX_BYTES_SIZE to keep this simple

algopy_testing/src/algopy_testing/contexts/state_context.py
```diff
@@ -0,0 +1,44 @@
+import threading
+from collections.abc import Generator
+from contextlib import contextmanager
+from typing import Any, Generic, TypeVar
+
+T = TypeVar("T")
+
+
+class TestContext(Generic[T]):
```
daniel-makerx, line 9: I think the contextvar module would be better here https://docs.python.org/3/library/contextvars.html

algopy_testing/src/algopy/op.py
```diff
@@ -0,0 +1,159 @@
...
+        return True
+
+
+def ed25519verify(a: Bytes | bytes, b: Bytes | bytes, c: Bytes | bytes, /) -> bool:
+    """
+    NOTE: Differs from AVM behavior since it does not having access to the precompiled
+    contract hash.
+    Currently works by forwarding the call to ed25519verify_bare
+    """
+    with state_context() as state:
```
daniel-makerx, line 67: If we use contextvar (see other comment) then this op should just try and access the current context with the error below if no current context is set

The testing code should be responsible for actually setting up the context using a `with` block

algopy_testing/src/algopy/op.py
```diff
@@ -0,0 +1,159 @@
...
+    if curve is None:
+        raise ValueError(f"Unsupported ECDSA curve: {v}")
+
+    try:
+        public_key = b"\x04" + pubkey_x_bytes + pubkey_y_bytes
+        vk = VerifyingKey.from_string(public_key, curve=curve)
+
+        # Concatenate R and S components to form the signature
+        signature = sig_r_bytes + sig_s_bytes
+        vk.verify_digest(signature, data_bytes)
```
daniel-makerx, line 110: Ideally try except blocks should be around just the lines that are expected to raise the exception

algopy_testing/tests/conftest.py
```diff
@@ -22,3 +28,25 @@ def algod_client() -> AlgodClient:
 @pytest.fixture(scope="session")
 def indexer_client() -> IndexerClient:
     return get_indexer_client(get_default_localnet_config("indexer"))
+
+
+@pytest.fixture(scope="module")
+def create_avm_invoker() -> typing.Callable[[ApplicationClient], AVMInvoker]:
```
daniel-makerx, line 34: might be simpler to just make this a helper function that takes an ApplicationClient and returns an AVMInvoker, rather than a fixture.

Then we can just have client specific fixtures as required

algopy_testing/src/algopy_testing/contexts/state_context.py
aorumbayev: Let me know what you think on the current approach in latest commit

algopy_testing/src/algopy/op.py
aorumbayev: @daniel-makerx let me know if the refinement makes sense, i do check on any context first but then it also checks on the specific program_hash field (which we set via context dictionary that user passes when state_context generator is invoked)

algopy_testing/src/algopy/op.py
```diff
@@ -0,0 +1,171 @@
...
+
+def keccak256(a: Bytes | bytes, /) -> Bytes:
+    input_value = as_bytes(a)
+    hashed_value = keccak.new(data=input_value, digest_bits=256)
+    return Bytes(hashed_value.digest())
+
+
+def sha512_256(a: Bytes | bytes, /) -> Bytes:
+    input_value = as_bytes(a)
+    hash_object = hashlib.new("sha512_256", input_value)
```
achidlow, line 46: this isn't cross-platform, it relies on OpenSSL (see CPython #71021)

since it looks like there's already a dependency, on Cryptodome, you can use that instead for cross-platform support:

```python3
from Cryptodome.Hash import SHA512

sha = SHA512.new(truncate="256")
sha.update(input_value)
return sha.digest()
```

daniel-makerx:
Ready to merge this now, one other thing that occurred to me is we should look at using [hatch's optional dependencies](https://hatch.pypa.io/dev/config/dependency/#features,) for the crypto dependencies, so they can be opted into with the extras syntax.

But that can be resolved in a seperate PR

daniel-makerx (approved):

PR #218, opened 2024-05-15
feat: BigUInt stub implementation

boblat:
## Proposed Changes

- add stub implementation of BigUInt which is backed by `bytes`
- when constructing `BigUInt` with `int` value (or `bytes` representation of it) larger than `2**512 - 1`,
  - a new `BigUInt` instance should be created successfully
  - accessing `.bytes` or `.value` properties should not throw error
  - using such `BigUInt` instance as input to an operation e.g. `+`, `&`, etc. should throw error before the operation is performed
- operations resulting in `BigUInt` value larger than `2**512 - 1` should **not** throw error
- operations resulting negative value should throw error
- bitwise operations (`&`, `|`, `^`) with UInt64 should return `BigUInt` value where the `.bytes` property should be padded with leading zeros (`b"\x00"`) so that the total length is always 8.


algopy_testing/src/algopy/primitives/biguint.py
```diff
@@ -0,0 +1,135 @@
...
+    from algopy.primitives.uint64 import UInt64
+
+# TypeError, ValueError are used for operations that are compile time errors
+# ArithmeticError and subclasses are used for operations that would fail during AVM execution
+
+
+@functools.total_ordering
+class BigUInt:
+    """
+    A python implementation of an AVM 64-bit unsigned integer
```
daniel-makerx, line 19: Should either remove this (as end users will normally only see the stub documentation) or update the description to reflect biguint (512-bit unsigned integer)

algopy_testing/src/algopy/primitives/biguint.py
```diff
@@ -0,0 +1,135 @@
...
+        return self.value < _as_int512(other)
+
+    def __bool__(self) -> bool:
+        return self.value != 0
+
+    def __add__(self, other: BigUInt | UInt64 | int) -> BigUInt:
+        return _checked_result(self.value + _as_int512(other), "+")
+
+    # the reflected dunder methods (__radd__, __rsub___, etc.) should only be called when the
+    # BigUInt is on the right hand side and an int on the left
```
daniel-makerx, line 50: For BigUInt we can expected slightly different behaviour, when calling a reflected method, the left hand side could be an `int` or a `UInt64`, so we should also use `_as_int512` to convert other

daniel-makerx: Actually I can see the implementation is doing that correctly, so just need to update this comment to avoid confusion

algopy_testing/tests/primitives/test_biguint.py
```diff
@@ -0,0 +1,449 @@
...
+        "le",
+        "gt",
+        "ge",
+    ],
+)
+def test_biguint_comparison(get_avm_result: AVMInvoker, op_name: str, a: int, b: int) -> None:
+    avm_result = get_avm_result(f"verify_biguint_{op_name}", a=a, b=b)
+    op = getattr(operator, op_name)
+
+    assert avm_result == op(BigUInt(a), BigUInt(b))
```
daniel-makerx, line 79: Most of these operations should also have some tests that compare mixed operands types with `BigUint` and `UInt64` 

e.g `BigUint(a) < UInt64(b)` and `UInt64(a) < BigUint(b)`

algopy_testing/src/algopy/primitives/biguint.py
boblat: It has been updated.

algopy_testing/src/algopy/primitives/biguint.py
boblat: I've updated the comment. I hope it is correct now. :)

algopy_testing/tests/primitives/test_biguint.py
boblat: I've added those tests.

algopy_testing/src/algopy/primitives/uint64.py
```diff
@@ -46,7 +50,7 @@ def __lt__(self, other: int | UInt64) -> bool:
     def __bool__(self) -> bool:
         return self.value != 0
 
-    def __add__(self, other: int | UInt64) -> UInt64:
+    def __add__(self, other: int | UInt64 | BigUInt) -> UInt64:
```
daniel-makerx, line 53: To resolve the type checking errors, we should instead update `BigUInt.__radd__` method (and other applicable dunder methods) to include `UInt64` in it's type hints. This will ensure our implementation and the stubs match

algopy_testing/src/algopy/primitives/biguint.py
daniel-makerx: Comment looks good :)

algopy_testing/tests/primitives/test_biguint.py
```diff
@@ -0,0 +1,701 @@
...
+        "ge",
+    ],
+)
+def test_biguint_comparison_uint64(
+    get_avm_result: AVMInvoker, op_name: str, a: int, b: UInt64
+) -> None:
+    avm_result = _get_avm_result(get_avm_result, f"verify_biguint_{op_name}_uint64", a, b)
+    op = getattr(operator, op_name)
+    assert avm_result == op(BigUInt(a), b)
+    if a < MAX_UINT64:
```
daniel-makerx, line 131: MAX_UINT64 is inclusive
```suggestion
    if a <= MAX_UINT64:
```

algopy_testing/src/algopy/primitives/biguint.py
```diff
@@ -0,0 +1,163 @@
...
+        return _as_int512(self) == _as_int512(other)
+
+    def __lt__(self, other: BigUInt | UInt64 | int) -> bool:
+        return _as_int512(self) < _as_int512(other)
+
+    def __bool__(self) -> bool:
+        return bool(_as_int512(self))
+
+    def __add__(self, other: BigUInt | UInt64 | int) -> BigUInt:
+        return _checked_result(_as_int512(self) + _as_int512(other), "+")
```
daniel-makerx, line 48: In order to match AVM behaviour, I think we need to remove the `_checked_result` call here so that an overflow in the result does not cause an error.

Instead an error should occur if the inputs to an arithmetic function have been exceeded i.e. if `_as_int512` returns a value > MAX_UINT512

algopy_testing/tests/primitives/test_biguint.py
```diff
@@ -0,0 +1,740 @@
...
+    [
+        "eq",
+        "ne",
+        "lt",
+        "le",
+        "gt",
+        "ge",
+    ],
+)
+def test_biguint_comparison(get_avm_result: AVMInvoker, op_name: str, a: int, b: int) -> None:
```
daniel-makerx, line 89: Since having a BigUInt with a value > MAX_UINT512 is possible, we should add tests to cover these cases for the comparison operations

algopy_testing/tests/primitives/test_biguint.py
```diff
@@ -0,0 +1,740 @@
...
+@pytest.mark.parametrize(
+    ("a", "b"),
+    [
+        (MAX_UINT512, 1),
+        (1, MAX_UINT512),
+        (MAX_UINT512, MAX_UINT512),
+    ],
+)
+def test_biguint_addition_result_overflow(get_avm_result: AVMInvoker, a: int, b: int) -> None:
+    avm_result = _get_avm_result(get_avm_result, "verify_biguint_add", a, b)[1]
```
daniel-makerx, line 186: See my earlier comment on the implementation, but this test is revealing that currently our behaviour does not match AVM since AVM does not error on producing an overflow result, but our implementation does.

algopy_testing/src/algopy/primitives/biguint.py
```diff
@@ -0,0 +1,163 @@
...
+        Returns:
+            BigUInt: The result of the unary positive operation.
+        """
+        return BigUInt(+_as_int512(self))
+
+    @classmethod
+    def from_bytes(cls, value: Bytes | bytes) -> BigUInt:
+        """Construct an instance from the underlying bytes (no validation)"""
+        result = cls()
+        result.__value = value.value if (isinstance(value, Bytes)) else value  # noqa: SLF001
```
daniel-makerx, line 116: Use `as_bytes` here so the correct validation is used

algopy_testing/src/algopy/primitives/biguint.py
boblat: Thanks Dan. It has been fixed in [this commit](https://github.com/algorandfoundation/puya/pull/218/commits/51d0a443d27076504a0b1eb3c2fb62a54696311c).

daniel-makerx (approved):

PR #225, opened 2024-05-20
feat: arithmetic op codes and miscellaneous op codes stub implementation

boblat:
## Proposed Changes

add stub implementation for the following arithmetic op codes and miscellaneous op codes

- [x] addw
- [x] base64_decode
- [x] bitlen
- [x] bsqrt
- [x] btoi
- [x] bzero
- [x] concat
- [x] divmodw
- [x] divw
- [x] err
- [x] exp
- [x] expw
- [x] extract
- [x] extract_uint16
- [x] extract_uint32
- [x] extract_uint64
- [x] getbit
- [x] getbytes
- [x] itob
- [x] json_ref
*Note*: The stub implementation of the `json_ref JSONObject` opcode returns a `[]bytes` representation of the object, with the following modifications:
  - All whitespace characters are removed.
  - All hexadecimal values are converted to lowercase.  
For instance, given a JSON payload like `b'{"key0": {"key1":    "\\uD801\\uDC37"}}'` and the key value `b"key0"`, the AVM would return `b'{"key1":    "\\uD801\\uDC37"}'`. However, the stub implementation would return `b'{"key1":"\\ud801\\udc37"}'`."
- [x] mulw
- [x] replace
- [x] select_bytes
- [x] select_uint64
- [x] setbit_bytes
- [x] setbit_uint64
- [x] setbyte
- [x] shl
- [x] shr
- [x] sqrt
- [x] substring


algopy_testing/tests/test_miscellaneous_op.py
```diff
@@ -0,0 +1,162 @@
...
+) -> None:
+    avm_result = get_ops_avm_result(
+        "verify_concat", a=a, b=b, pad_a_size=pad_a_size, pad_b_size=pad_b_size
+    )
+    a = (b"\x00" * pad_a_size) + a
+    b = (b"\x00" * pad_b_size) + b
+    assert avm_result == get_sha256_hash(op.concat(a, b))
+
+
+def get_sha256_hash(v: Bytes) -> Bytes:
```
aorumbayev, line 157: should this be private ? I assume its not used outside of this file?

boblat: It is now using the implementation in `util.py` as it is the same.

algopy_testing/src/algopy/op.py
```diff
@@ -152,3 +154,46 @@ def ecdsa_pk_decompress(v: ECDSA, a: Bytes | bytes, /) -> tuple[Bytes, Bytes]:
         return Bytes(pubkey_x.to_bytes(32, byteorder="big")), Bytes(
             pubkey_y.to_bytes(32, byteorder="big")
         )
+
+
+def addw(a: UInt64 | int, b: UInt64 | int, /) -> tuple[UInt64, UInt64]:
+    a = as_int(a, max=MAX_UINT64)
```
daniel-makerx, line 160: There is an `as_int64` method in utils now that does this

algopy_testing/src/algopy/op.py
```diff
@@ -152,3 +154,46 @@ def ecdsa_pk_decompress(v: ECDSA, a: Bytes | bytes, /) -> tuple[Bytes, Bytes]:
...
+    int_value = (
+        int.from_bytes(as_bytes(a))
+        if (isinstance(a, Bytes | bytes))
+        else as_int(a, max=MAX_UINT64)
+    )
+    return UInt64(int_value.bit_length())
+
+
+def bsqrt(a: BigUInt | int, /) -> BigUInt:
+    a = as_int(a, max=MAX_UINT512)
```
daniel-makerx, line 180: Can use the new `as_int512` here also

algopy_testing/src/algopy/op.py
```diff
@@ -152,3 +154,46 @@ def ecdsa_pk_decompress(v: ECDSA, a: Bytes | bytes, /) -> tuple[Bytes, Bytes]:
...
+    return BigUInt(math.isqrt(a))
+
+
+def btoi(a: Bytes | bytes, /) -> UInt64:
+    a_bytes = as_bytes(a)
+    if len(a_bytes) > 8:
+        raise AssertionError(f"arg too long, got [{len(a_bytes)}]bytes")
```
daniel-makerx, line 187: For consistency I think this should be a `ValueError`

algopy_testing/tests/common.py
```diff
@@ -22,7 +22,9 @@ def invoke(method: str, **kwargs: typing.Any) -> object:
             },
             **kwargs,
         ).return_value
-        if isinstance(result, list) and all(isinstance(i, int) for i in result):
+        if isinstance(result, list) and all(
+            isinstance(i, int) and i >= 0 and i <= 256 for i in result
```
daniel-makerx, line 26:
```suggestion
            isinstance(i, int) and i >= 0 and i <= 255 for i in result
```

algopy_testing/src/algopy/op.py
```diff
@@ -168,3 +178,398 @@ def vrf_verify(
...
+
+
+def divw(a: UInt64 | int, b: UInt64 | int, c: UInt64 | int, /) -> UInt64:
+    i = _uint128_to_int(a, b)
+    c = as_int64(c)
+    return UInt64(i // c)
+
+
+def err() -> None:
+    raise ArithmeticError("err opcode executed")
```
daniel-makerx, line 241: I think ArithmeticError should only be used for arithmetic like operations, can use RuntimeError here instead

algopy_testing/src/algopy/op.py
```diff
@@ -168,3 +178,398 @@ def vrf_verify(
...
+    result = a**b
+    return _int_to_uint128(result)
+
+
+def extract(a: Bytes | bytes, b: UInt64 | int, c: UInt64 | int, /) -> Bytes:
+    a = a if (isinstance(a, Bytes)) else Bytes(a)
+    b = as_int64(b)
+    c = as_int64(c)
+    if b > len(a):
+        raise ArithmeticError(f"extraction start {b} is beyond length")
```
daniel-makerx, line 266: Use ValueError here

algopy_testing/src/algopy/op.py
```diff
@@ -168,3 +178,398 @@ def vrf_verify(
...
+def expw(a: UInt64 | int, b: UInt64 | int, /) -> tuple[UInt64, UInt64]:
+    a = as_int64(a)
+    b = as_int64(b)
+    if a == b and a == 0:
+        raise ArithmeticError("0^0 is undefined")
+    result = a**b
+    return _int_to_uint128(result)
+
+
+def extract(a: Bytes | bytes, b: UInt64 | int, c: UInt64 | int, /) -> Bytes:
```
daniel-makerx, line 261: There is a special edge case we need to handle here: 
When b and c are are both `int` (and only `int`) and b <= 255, and c == `0` then the c is treated as to the end of the bytes array.

See https://developer.algorand.org/docs/get-details/dapps/avm/teal/opcodes/v10/#extract

algopy_testing/src/algopy/op.py
```diff
@@ -168,3 +178,398 @@ def vrf_verify(
...
+    return UInt64(result_int)
+
+
+def getbit(a: Bytes | UInt64 | bytes | int, b: UInt64 | int, /) -> UInt64:
+    if isinstance(a, Bytes | bytes):
+        return _getbit_bytes(a, b)
+    if isinstance(a, UInt64 | int):
+        a_bytes = _uint64_to_bytes(a)
+        return _getbit_bytes(a_bytes, b, "little")
+    raise ValueError("Unknown type for argument a")
```
daniel-makerx, line 296: Use TypeError here

algopy_testing/src/algopy/op.py
```diff
@@ -168,3 +178,398 @@ def vrf_verify(
...
+    return UInt64(result)
+
+
+def sqrt(a: UInt64 | int, /) -> UInt64:
+    a = as_int64(a)
+    return UInt64(math.isqrt(a))
+
+
+def substring(a: Bytes | bytes, b: UInt64 | int, c: UInt64 | int, /) -> Bytes:
+    a = a if (isinstance(a, Bytes)) else Bytes(a)
```
daniel-makerx, line 385: minor: using as_bytes here and then converting to `Bytes` at the end might be easier to read?

algopy_testing/src/algopy/op.py
```diff
@@ -168,3 +178,398 @@ def vrf_verify(
...
+
+def _uint128_to_int(a: UInt64 | int, b: UInt64 | int) -> int:
+    a = as_int64(a)
+    b = as_int64(b)
+    return (a << 64) + b
+
+
+def _uint64_to_bytes(a: UInt64 | int) -> bytes:
+    a = as_int64(a)
+    a_bytes = _int_list_to_bytes([a])
```
daniel-makerx, line 413: could this just be `a.to_bytes(length=8)` ?

algopy_testing/src/algopy/op.py
```diff
@@ -168,3 +178,398 @@ def vrf_verify(
...
+            raise TypeError(f"value must be an object type, not {type(result).__name__!r}")
+
+        result = _MultiKeyDict(result)
+        result_string = json.dumps(result, separators=(",", ":"))
+        return Bytes(result_string.encode())
+
+
+class _MultiKeyDict(dict[Any, Any]):
+    def __init__(self, items: list[Any]):
+        self["something"] = "something"
```
daniel-makerx, line 563: Is this meant to be here?

algopy_testing/src/algopy/op.py
```diff
@@ -168,3 +178,398 @@ def vrf_verify(
...
+            # all other levels remain as key value pairs
+            # e.g.
+            # input bytes: b'{"key0": 1,"key1": {"key2":2,"key2":"10"}, "key2": "test"}'
+            # output dict: {'key0': 1, 'key1': [('key2', 2), ('key2', '10')], 'key2': 'test'}
+            result = dict(pairs)
+            if len(pairs) != len(result):
+                raise ValueError(
+                    "error while parsing JSON text, invalid json text, duplicate keys found"
+                )
+            return result  # noqa: TRY300
```
daniel-makerx, line 505: Instead of suppressing this lint error, can modify the try block to just be around the json.loads statement

algopy_testing/src/algopy/op.py
boblat: It has been fixed in [this commit](bcae2a50a453378b1c97e2fbd94600629be6c12e)

algopy_testing/src/algopy/op.py
boblat: It has been fix in the commit 13e6945f9666d636f08357e178ae75268ceda393.

algopy_testing/src/algopy/op.py
boblat: It seems that `op.extract` is compiled down to `extract3` in teal which does not seem to behave the same as `extract` op code when `c == 0`. In the tests, I got empty string for the test case where both `a` and `b` are `0`. 

https://github.com/algorandfoundation/puya/pull/225/files#diff-0457a536f94ff2f407b843b52ba75b1e22c2482c4714d009cd2dc7a9156c45f1R585 

https://developer.algorand.org/docs/get-details/dapps/avm/teal/opcodes/v10/#extract3

algopy_testing/src/algopy/op.py
boblat: It is fixed in the commit be7f752396a4563859209c063f7a03577a4f680f.

algopy_testing/src/algopy/op.py
boblat: It is fixed in the commit 424cb6c20b396b310e5a68a83966f03e28642d89.

algopy_testing/src/algopy/op.py
daniel-makerx: To replicate it in puya you will need to use int literals not UInt64,  e.g.

`
op.extract(some_bytes, 5, 0)
`

Probably will need a seperate method for this edge case

algopy_testing/src/algopy/op.py
boblat: Nice. It has been fixed in this commit 7ec1e0f7b9b4b9fd5d85bc5bec9460d3ac58253d.

algopy_testing/src/algopy/op.py
boblat: I got it off [this stackoverflow answer](https://stackoverflow.com/a/29520802). Without that line, we get an empty object when serialising.

algopy_testing/src/algopy/op.py
boblat: Some tests for the edge case are added in this commit 267f745e35d81a35e41fe6ab2376ece5a4ce953a.

algopy_testing/tests/test_miscellaneous_op.py
```diff
@@ -0,0 +1,1581 @@
...
+    result = op.addw(a, b)
+    assert avm_result_tuples == result
+
+
+@pytest.mark.parametrize(
+    ("a", "b"),
+    [(1, MAX_UINT64 + 1), (MAX_UINT64 + 1, 1), (0, MAX_UINT512), (MAX_UINT512 * 2, 0)],
+)
+def test_addw_input_overflow(get_ops_avm_result: AVMInvoker, a: int, b: int) -> None:
+    with pytest.raises(ABIEncodingError, match=_avm_int_arg_overflow_error):
```
daniel-makerx, line 91: This particular error is coming from the SDK, not the AVM so I think we can just remove it and just have the testing implementation check below

algopy_testing/tests/test_miscellaneous_op.py
```diff
@@ -0,0 +1,1581 @@
...
+    result = op.bitlen(a)
+    assert avm_result == result
+
+
+@pytest.mark.parametrize(
+    "a",
+    [MAX_UINT64 + 1, MAX_UINT512, MAX_UINT512 * 2],
+)
+def test_uint64_bitlen_input_overflow(get_ops_avm_result: AVMInvoker, a: int) -> None:
+    with pytest.raises(ABIEncodingError, match=_avm_int_arg_overflow_error):
```
daniel-makerx, line 197: Likewise can remove this check since it is an SDK error

algopy_testing/tests/test_miscellaneous_op.py
```diff
@@ -0,0 +1,1581 @@
...
+    with pytest.raises(ValueError, match="bzero attempted to create a too large string"):
+        op.bzero(a)
+
+
+@pytest.mark.parametrize(
+    "a",
+    [MAX_UINT64 + 1, MAX_UINT512, MAX_UINT512 * 2],
+)
+def test_bzero_input_overflow(get_ops_avm_result: AVMInvoker, a: int) -> None:
+    with pytest.raises(ABIEncodingError, match=_avm_int_arg_overflow_error):
```
daniel-makerx, line 291: Another ABI encoding error

daniel-makerx (approved):

algopy_testing/tests/test_miscellaneous_op.py
boblat: All tests for ABIEncodingError have been removed in this commit 414af4e35198510db081551ed9370a944178090c.

PR #226, opened 2024-05-20
feat: String primitive implementation

aorumbayev:
https://github.com/algorandfoundation/algokit-internal/issues/101

## Proposed Changes

- Adds string implementation, largely mirroring the Byte excluding the len, slicing and indexing support


aorumbayev:
@achidlow @daniel-makerx 


## Open questions

Getting a weird 

```
tests/artifacts/PrimitiveOps/contract.py:347 error: Argument 1 to "Bytes" has incompatible type "Bytes"; expected "bytes"  [arg-type]
                                                               return Bytes(result.bytes)
                                                                            ^~~~~~~~~~~~
Error: Command '['hatch', 'run', 'puyapy', 'tests/artifacts/PrimitiveOps/contract.py', '--out-dir', 'data']' returned non-zero exit status 1.
```

error when trying to compile a contract to test out the String join behaviour. 

```python
    @arc4.abimethod
    def verify_string_join(self, a: String, b: String) -> String:
        result = String(", ").join((a, b))
        return result

=== OR ====

    @arc4.abimethod
    def verify_string_join(self, a: String, b: String) -> Bytes:
        result = String(", ").join((a, b))
        return Bytes(result.bytes)
```

Have you seen any similar issues, it seems like it identifies the joined value as Bytes instead of a String? Also to be clear this String is going to differ from abi.String, was this created specifically for algorand python to mirror the str in python?

algopy_testing/src/algopy/utils.py
```diff
@@ -59,3 +59,19 @@ def as_bytes(value: object, *, max_size: int = MAX_BYTES_SIZE) -> bytes:
...
+    from algopy import String
+
+    match value:
+        case str(string_value):
+            pass
+        case String(value=string_value):
+            pass
+        case _:
+            raise TypeError(f"value must be a string or String type, not {type(value).__name__!r}")
+    if len(string_value) > MAX_STRING_SIZE:
```
daniel-makerx, line 74: Since UTF8 can contain code points consisting of multiple bytes, we can't check the string length, instead need to check the length of the underlying bytes e.g.
```suggestion
    if len(string_value.encode("utf8")) > MAX_BYTES_SIZE:
```

algopy_testing/src/algopy/primitives/string.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def starts_with(self, prefix: String | str) -> bool:
+        return self.value.startswith(as_string(prefix))
+
+    def ends_with(self, suffix: String | str) -> bool:
+        return self.value.endswith(as_string(suffix))
+
+    def join(self, others: tuple[String, ...], /) -> String:
+        byte_values = [as_string(s) for s in others]
+        return String(f"{self.value}".join(byte_values))
```
daniel-makerx, line 51:
```suggestion
        return String(self.value.join(map(as_string, others)))
```

daniel-makerx:
> Have you seen any similar issues, it seems like it identifies the joined value as Bytes instead of a String? 
This is a bug, I'll get a fix into main for it

> Also to be clear this String is going to differ from abi.String, was this created specifically for algorand python to mirror the str in python?
Yes it is analogous to the builtin str type, main difference between this and arc4.String is arc4.String is a dynamic array so by arc4 encoding rules it includes the length of the string as a header

algopy_testing/src/algopy/primitives/string.py
```diff
@@ -0,0 +1,54 @@
...
+    def __bool__(self) -> bool:
+        return bool(self.value)
+
+    def __eq__(self, other: object) -> bool:
+        return self.value == as_string(other)
+
+    def __contains__(self, item: object) -> bool:
+        return as_string(item) in self.value
+
+    def __len__(self) -> int:
```
daniel-makerx, line 34: `__len__` isn't in the stubs, so should be able to just omit this?

algopy_testing/src/algopy_testing/constants.py
```diff
@@ -1,6 +1,8 @@
 UINT64_SIZE = 64
 UINT512_SIZE = 512
 MAX_BYTES_SIZE = 4096
+MAX_STRING_SIZE = 4096
```
daniel-makerx, line 4: I don't think we want a seperate size for String, as they have the same limitation as the bytes type in terms of size, it's just the Stubs have them as a distinct type

algopy_testing/src/algopy/utils.py
```diff
@@ -59,3 +59,19 @@ def as_bytes(value: object, *, max_size: int = MAX_BYTES_SIZE) -> bytes:
     if len(bytes_value) > max_size:
         raise ValueError(f"expected value length <= {max_size}, got: {len(bytes_value)}")
     return bytes_value
+
+
+def as_string(value: object, *, max_size: int = MAX_STRING_SIZE) -> str:
```
daniel-makerx, line 64: I think we can remove the max_size parameter also, it's not used below and I don't think we will have need of having different sized String's beyond the normal max

daniel-makerx (approved):

PR #228, opened 2024-05-20
feat: vrf_verify mock

aorumbayev:
https://github.com/algorandfoundation/algokit-internal/issues/20

## Proposed Changes

- Adding verify_vrf method to be used for mocking purposes.


aorumbayev:
@daniel-makerx @achidlow , the main implementation of vrf is in go-algorand repo and its in go, there are (probably research oriented) snippets of vrf implementation in python but its 6 years old and has python2 dependency available on https://github.com/algorand/vrf.

Given the following factors:
1. complexity of maintaining and implementing a newer version of vrf in python
2. functionality being aimed at testing contexts 

It seemed like implementing the vrf in python just for the testing scenarios seemed like an overkill, given that a user can always mock the behaviour. Let me know if you agree with keeping this without matching python implementation and recommending users to mock it instead?

daniel-makerx:
>Let me know if you agree with keeping this without matching python implementation and recommending users to mock it instead?

This seems reasonable, if a convenient python implementation is ever available we can provide an implementation then

algopy_testing/tests/test_op.py
```diff
@@ -314,3 +315,34 @@ def test_verify_ecdsa_decompress_k1(
...
+    )
+    c = bytes.fromhex("3a2740da7a0788ebb12a52154acbcca1813c128ca0b249e93f8eb6563fee418d")
+
+    def run_real_vrf_verify() -> tuple[Bytes, bool]:
+        sp = algod_client.suggested_params()
+        sp.fee = 6000
+        result = get_crypto_ops_avm_result("verify_vrf_verify", a=a, b=b, c=c, suggested_params=sp)
+        return (Bytes(bytes(result[0])), bool(result[1]))  # type: ignore  # noqa: PGH003
+
+    def run_mocked_vrf_verify() -> tuple[Bytes, bool]:
```
daniel-makerx, line 340: Is this done as a nested function because of the mocking?

algopy_testing/tests/test_op.py
```diff
@@ -314,3 +315,34 @@ def test_verify_ecdsa_decompress_k1(
...
+        return op.vrf_verify(op.VrfVerify.VrfAlgorand, a, b, c)
+
+    avm_result = run_real_vrf_verify()
+    mocker.patch("algopy.op.vrf_verify", return_value=[avm_result[0], True])
+    mocked_result = run_mocked_vrf_verify()
+
+    assert avm_result[0] == mocked_result[0]
```
daniel-makerx, line 347: Could this just be `aasert avm_result == mocked_result`?

algopy_testing/tests/test_op.py
aorumbayev: @daniel-makerx yes, figured a nested function is fine in this context cause its only needed within this test

algopy_testing/src/algopy/op.py
```diff
@@ -152,3 +156,15 @@ def ecdsa_pk_decompress(v: ECDSA, a: Bytes | bytes, /) -> tuple[Bytes, Bytes]:
         return Bytes(pubkey_x.to_bytes(32, byteorder="big")), Bytes(
             pubkey_y.to_bytes(32, byteorder="big")
         )
+
+
+def vrf_verify(
+    s: VrfVerify,  # noqa: ARG001
```
daniel-makerx, line 162: since these arguments are positional only, can prefix the arg names with _ and then remove the `# noqa`

daniel-makerx (approved):

algopy_testing/src/algopy/op.py
aorumbayev: @daniel-makerx this wouldn't be an issue with the fact that the argument names will be different from the ones in the stub definition? The '_' in front of args wouldn't affect anything?

robdmoore:
100% - just allow it to be mocked via a test double. The mental model I have here is per this 10 year old article: https://serialseb.com/blog/2013/07/11/unit-testing-is-out-vertical-slice-testing-is-in/ (noting that naturally anything we can substitute with the same implementation easily is obvious fantastic, but we have to draw pragmatic lines like you are doing here <3).

PR #233, opened 2024-05-27
fix: fix inner transaction validation to handle some cases that were accidentally missed

daniel-makerx:


src/puya/awst_build/validation/inner_transactions.py
```diff
@@ -11,16 +11,36 @@
 
 logger = log.get_logger(__name__)
 
+INNER_TRANSACTION_ASSIGNMENT_EXPRESSION_ERROR = (
+    "Inner Transactions cannot be used in assignment expressions."
```
achidlow, line 15: probably want to lower case `Inner Transaction(s)` in these messages, and remove the full stop at the end for single sentences?

PR #235, opened 2024-05-28
feat: baseline for state management abstractions

aorumbayev:
## Proposed Changes

- Initial draft implementation of the unit testing abstractions to aid unit testing of algorand python contracts

## Further notes

1. This is a draft, a detailed description for the PR is to follow 
2. The overall approach differs from @tristanmenzel 's approach in puya-ts, is semantic compatibility across languages even an issue/priority?


algopy_testing/examples/auction/test_contract.py
```diff
@@ -0,0 +1,146 @@
...
+def contract() -> AuctionContract:
+    return AuctionContract()
+
+
+def test_opt_into_asset(
+    context: TestContext[Any], contract: AuctionContract, mocker: MagicMock
+) -> None:
+    # Arrange
+    dummy_address = "A" * 58
+    context.global_state.creator_address = Account(dummy_address)
```
robdmoore, line 28: I think `global_state` is confusing and could be confused with global strorage. Let's call it `global`?

daniel-makerx: `global` is a keyword in python so wouldn't be able to use that `global_` would be valid

algopy_testing/examples/auction/test_contract.py
```diff
@@ -0,0 +1,146 @@
...
+    assert contract.claimable_amount[dummy_address] == 200
+
+
+def test_claim_bids(
+    context: TestContext[Any], contract: AuctionContract, mocker: MagicMock
+) -> None:
+    # Arrange
+    dummy_account = Account("A" * 58)
+    context.transaction_state.sender = dummy_account
+    contract.claimable_amount[dummy_account] = 300
```
daniel-makerx, line 91: I would expect this to fail type check with mypy when using the examples venv? It should be invalid to directly assign an `int` like this to local_state

algopy_testing/src/algopy_testing/models/account.py
```diff
@@ -0,0 +1,84 @@
...
+    def __bool__(self) -> bool:
+        return self._public_key != algosdk.constants.ZERO_ADDRESS
+
+    def __hash__(self) -> int:
+        return hash(self._public_key)
+
+    def is_opted_in(self, asset_or_app: Asset | Application, /) -> bool:
+        raise NotImplementedError(
+            "The 'is_opted_in' method is being executed in a python testing context. "
+            "Please mock this method in according to your python testing framework of choice."
```
neilcampbell, line 67:
```suggestion
            "Please mock this method using your python testing framework of choice."
```
Probably makes sense to define these messages as constants, so we can consistently apply the terminology.

algopy_testing/examples/auction/test_contract.py
```diff
@@ -0,0 +1,143 @@
...
+
+def test_opt_into_asset(
+    context: TestContext[Any],
+) -> None:
+    # Arrange
+    dummy_account = context.any_account()
+    context.set_global_fields(
+        {
+            "creator_address": dummy_account,
+            "current_application_address": dummy_account,
```
daniel-makerx, line 26: minor: this test should be using a different account for the application address

algopy_testing/src/algopy_testing/gtxn.py
```diff
@@ -0,0 +1,6 @@
+class AssetTransferTransaction:
```
daniel-makerx, line 1: At the moment this allows the mock to use invalid types, should make this have the appropriate fields so mocks will be type checked correctly. for now just add the fields required for the examples test with a TODO for the others

algopy_testing/src/algopy_testing/models/account.py
```diff
@@ -0,0 +1,84 @@
...
+    def __repr__(self) -> str:
+        return self._public_key
+
+    def __str__(self) -> str:
+        return self._public_key
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Account):
+            return self._public_key == other._public_key
+        return self._public_key == as_string(other)
```
daniel-makerx, line 53: only the `str` and `Account` types should be accepted, `String` should be a TypeError to match puya compile behaviour

algopy_testing/src/algopy_testing/models/account.py
```diff
@@ -0,0 +1,84 @@
...
+            )
+        if isinstance(value, bytes):
+            decoded_value = algosdk.encoding.encode_address(value)
+        else:
+            decoded_value = str(value)
+        return cls(decoded_value)
+
+    @property
+    def bytes(self) -> Bytes:
+        return Bytes(self._public_key.encode("utf-8"))
```
daniel-makerx, line 84: the underlying bytes should be the public key portion of the bytes after using algosdk.encoding.decode_address

algopy_testing/src/algopy_testing/models/global_state.py
```diff
@@ -0,0 +1,71 @@
...
+if TYPE_CHECKING:
+    from algopy_testing.models.account import Account
+    from algopy_testing.models.application import Application
+    from algopy_testing.primitives.bytes import Bytes
+    from algopy_testing.primitives.uint64 import UInt64
+
+T = TypeVar("T")
+
+
+class GlobalFieldsKwargs(TypedDict, total=False):
```
daniel-makerx, line 16: I think we should avoid having two ways of defining these values, tried this approach with algokit_utils but it ended up just creating more issues, so would want to avoid doing that again. My preference would be to just have the dataclass style (with default values) as this gives better IDE support and the only con is it's slightly more verbose

algopy_testing/src/algopy_testing/models/global_state.py
```diff
@@ -0,0 +1,71 @@
...
+    asset_opt_in_min_balance: UInt64 | None = None
+    genesis_hash: Bytes | None = None
+    opcode_budget: Callable[[], int] | None = None
+
+
+class _Global:
+    def __getattr__(self, name: str) -> object:
+        from algopy_testing.context import get_test_context
+
+        if name in GlobalFields.__annotations__:
```
daniel-makerx, line 62: Don't need to check if the name is in annotations, getattr will throw an AttributeError if the name is incorrect. 

Additionally, we should check the value returned by getattr is not None and throw if it is as this indicates the value has not been set/mocked but is being referenced somewhere. None is not a valid value for compiled contracts so we shouldn't allow stub implementations to return it as a value

algopy_testing/src/algopy_testing/models/global_state.py
aorumbayev: @daniel-makerx the problem with a single dataclass per abstraction in models (for things like global, txn, itxn) was the lack of straightforward way to allow users to have typed individual fields for each abstraction that they can set on the context via methods like patch_global_fields and etc. 

@daniel-makerx let me know what you think of the latest implementation. 

I am using a TypedDict (GlobalFields and etc) per each and then dynamically fetch the values in __getattr__ dunder in the dataclass, the exported object is an instance of the dataclass (given that the __getattr__ only gets invoked when accessing attributes on instances). 

As a last tweak, i am also using Unpack from `typing` to unpack the TypedDict as a **kwargs field a user can pass on the add_* methods. That way a user gets to be able to easily patch any field on the Global, Txn and ITxn - and then within mocked contract execution accessing the values like Global will in turn access an instance of a dataclass where the getattr dunder will take care of loading the patched field (if exists) from the dictionary set on the global context instance. 

I believe this still addresses the maintainability issue of having the same field duplicated on typeddict vs dataclass

algopy_testing/src/algopy/__init__.py
```diff
@@ -1,5 +1,101 @@
...
+    ) -> None:
+        self.type_ = type_
+        self.key = key
+        self.description = description
+        self._state: dict[object, object] = {}
+
+    def __setitem__(self, key: object, value: object) -> None:
+        if isinstance(key, Account):
+            key = str(key)
+        elif isinstance(key, Application | Asset):
```
daniel-makerx, line 35: This doesn't look right. local state allows indexing by UInt64 or int, which represents the foreign account index to use (so ultimately still indexing by Account). Any other types should be a TypeError

aorumbayev: @daniel-makerx good catch, added a validation to only accept account, int or uint64

algopy_testing/src/algopy_testing/context.py
```diff
@@ -1,24 +1,495 @@
...
+    import algopy
+    import algopy.gtxn
+
+    from algopy_testing.itxn import BaseInnerTransaction
+
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+class AlgopyTestContext(Generic[T]):
```
daniel-makerx, line 30: I don't think this need to be generic anymore?

algopy_testing/src/algopy_testing/context.py
```diff
@@ -1,24 +1,495 @@
...
+    def __init__(self, contract_cls: type[T] | None = None) -> None:
```
daniel-makerx, line 31: I think this can be simplified as well
```suggestion
    def __init__(self, contract_cls: type[algopy.Contract] | type[algopy.ARC4Contract] | None = None) -> None:
```

algopy_testing/src/algopy_testing/context.py
```diff
@@ -1,24 +1,495 @@
...
+        self._contract_cls = (
+            contract_cls  # TODO: not utilized yet, just storing the reference for now
+        )
+        self.account_data: dict[str, AccountData] = {}
+        self.application_data: dict[int, ApplicationFields] = {}
+        self.asset_data: dict[int, AssetFields] = {}
+        self.inner_transactions: list[BaseInnerTransaction] = []
+        self.global_fields: dict[str, Any] = {}
```
daniel-makerx, line 39: These properties can use the typed dicts now?

algopy_testing/src/algopy_testing/models/account.py
```diff
@@ -0,0 +1,157 @@
...
+
+    @classmethod
+    def from_bytes(cls, value: algopy.Bytes | bytes) -> Self:
+        from algopy import Bytes
+
+        if not isinstance(value, (Bytes | bytes)):
+            raise TypeError("Invalid value for Account")
+
+        value_len = value.length if isinstance(value, Bytes) else len(value)
+        if value_len != 32:
```
daniel-makerx, line 79: Puya's implementation `from_bytes` does not do any validation beyond type, so to match the compiled behaviour this should not check the length or try to encode it

algopy_testing/src/algopy_testing/models/account.py
```diff
@@ -0,0 +1,157 @@
...
+        raise TypeError(
+            "Invalid `asset_or_app` argument type. Must be an `algopy.Asset` or "
+            "`algopy.Application` instance."
+        )
+
+    @classmethod
+    def from_bytes(cls, value: algopy.Bytes | bytes) -> Self:
+        from algopy import Bytes
+
+        if not isinstance(value, (Bytes | bytes)):
```
daniel-makerx, line 75: Use the `as_bytes` util here

algopy_testing/src/algopy_testing/models/account.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+@dataclass()
+class Account:
+    _public_key: str
+
+    def __init__(self, value: str | Bytes = algosdk.constants.ZERO_ADDRESS, /):
+        if not isinstance(value, (str | Bytes)):
+            raise TypeError("Invalid value for Account")
+        if isinstance(value, Bytes):
```
daniel-makerx, line 40: I think the underlying value stored here should be the bytes, and convert to the human readable form as required. However we should check the bytes is of length 32 to match puya.

algopy_testing/src/algopy_testing/models/account.py
```diff
@@ -0,0 +1,157 @@
...
+        return self._public_key
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Account | str):
+            raise TypeError("Invalid value for Account")
+        if isinstance(other, Account):
+            return self._public_key == other._public_key
+        return self._public_key == as_string(other)
+
+    def __ne__(self, other: object) -> bool:
```
daniel-makerx, line 132: This can be omitted, the default implementation will use the inverse of __eq__

algopy_testing/src/algopy_testing/models/application.py
```diff
@@ -0,0 +1,74 @@
...
+            )
+
+        return return_value
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Application):
+            return self.id == other.id
+        return self.id == as_string(other)
+
+    def __ne__(self, other: object) -> bool:
```
daniel-makerx, line 67: Can omit

algopy_testing/src/algopy_testing/models/account.py
```diff
@@ -0,0 +1,157 @@
...
+        if self._public_key not in context.account_data:
+            raise ValueError(
+                "`algopy.Account` is not present in the test context! "
+                "Use `context.add_account()` or `context.any_account()` to add the account "
+                "to your test setup."
+            )
+
+        return_value = context.account_data[self._public_key].fields.get(name)
+        if return_value is None:
+            raise TypeError(
```
daniel-makerx, line 111: I think this should be a ValueError too?

algopy_testing/src/algopy_testing/models/application.py
```diff
@@ -0,0 +1,74 @@
...
+        if int(self.id) not in context.application_data:
+            raise ValueError(
+                "`algopy.Application` is not present in the test context! "
+                "Use `context.add_application()` or `context.any_application()` to add the "
+                "application to your test setup."
+            )
+
+        return_value = context.application_data[int(self.id)].get(name)
+        if return_value is None:
+            raise TypeError(
```
daniel-makerx, line 54: ValueError

algopy_testing/src/algopy_testing/models/asset.py
```diff
@@ -0,0 +1,112 @@
...
+        if int(self.id) not in context.asset_data:
+            raise ValueError(
+                "`algopy.Asset` is not present in the test context! "
+                "Use `context.add_asset()` or `context.any_asset()` to add the asset to "
+                "your test setup."
+            )
+
+        return_value = context.asset_data[int(self.id)].get(name)
+        if return_value is None:
+            raise TypeError(
```
daniel-makerx, line 92: ValueError

algopy_testing/src/algopy_testing/models/asset.py
```diff
@@ -0,0 +1,112 @@
...
+            )
+
+        return return_value
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Asset):
+            return self.id == other.id
+        return self.id == other
+
+    def __ne__(self, other: object) -> bool:
```
daniel-makerx, line 105: Can Omit

algopy_testing/src/algopy_testing/models/global_state.py
```diff
@@ -0,0 +1,67 @@
...
+                "Use `context.patch_global_fields()` to set the fields in your test setup."
+            )
+        if name not in context.global_fields:
+            raise AttributeError(
+                f"'algopy.Global' object has no value set for attribute named '{name}'. "
+                f"Use `context.patch_global_fields({name}=your_value)` to set the value "
+                "in your test setup."
+            )
+        if not context.global_fields[name]:
+            raise TypeError(
```
daniel-makerx, line 58: AttributeError

algopy_testing/src/algopy_testing/models/itxn.py
```diff
@@ -0,0 +1,101 @@
...
+                "Use `context.patch_itxn_fields()` to set the fields in your test setup."
+            )
+        if name not in context.itxn_fields:
+            raise AttributeError(
+                f"'Txn' object has no value set for attribute named '{name}'. "
+                f"Use `context.patch_itxn_fields({name}=your_value)` to set the value "
+                "in your test setup."
+            )
+        if not context.itxn_fields[name]:
+            raise TypeError(
```
daniel-makerx, line 92: AttributeError

algopy_testing/src/algopy_testing/models/txn.py
```diff
@@ -0,0 +1,115 @@
...
+                "`algopy.Txn` fields are not set in the test context! "
+                "Use `context.patch_txn_fields()` to set the fields in your test setup."
+            )
+        if name not in context.txn_fields:
+            raise AttributeError(
+                f"'Txn' object has no value set for attribute named '{name}'. "
+                f"Use `context.patch_txn_fields({name}=your_value)` to set the value "
+                "in your test setup."
+            )
+        if not context.txn_fields[name]:
```
daniel-makerx, line 105: Not required any more, see other comment

algopy_testing/src/algopy_testing/context.py
```diff
@@ -1,24 +1,495 @@
...
+        self.txn_fields: dict[str, Any] = {}
+        self.itxn_fields: dict[str, Any] = {}
+        self.gtxns: list[algopy.gtxn.TransactionBase] = []
+        self.logs: list[str] = []
+        self._asset_id_count: int = 1
```
daniel-makerx, line 44: Can use this trick to generate valid id's
```suggestion
        self._asset_id = iter(range(1001, 2**64))
```
Then when you need the next id it is just `next(self._asset_id)`

algopy_testing/src/algopy_testing/models/itxn.py
```diff
@@ -0,0 +1,101 @@
...
+    def __getattr__(self, name: str) -> object:
+        from algopy_testing.context import get_test_context
+
+        context = get_test_context()
+        if not context:
+            raise ValueError(
+                "Test context is not initialized! Use `with algopy_testing_context()` to access "
+                "the context manager."
+            )
+        if not context.itxn_fields:
```
daniel-makerx, line 80: rather than having a seperate type dict, this can just look at the last inner transaction e.g. `context.inner_transactions[-1]`

PR #236, opened 2024-05-30
feat: stub implementation of arc4 UIntN and BigUIntN

boblat:
## Proposed Changes
Implement stubs for 
 - UIntN
 - BigUIntN
 - UInt8
 - UInt16
 - UInt32
 - UInt64
 - UInt128
 - UInt256
 - UInt512
 - Byte


algopy_testing/src/algopy_testing/utils.py
```diff
@@ -76,5 +81,10 @@ def as_string(value: object) -> str:
             raise TypeError(f"value must be a string or String type, not {type(value).__name__!r}")
 
 
-def int_to_bytes(x: int) -> bytes:
-    return x.to_bytes((x.bit_length() + 7) // 8, "big")
+def int_to_bytes(x: int, pad_to: int | None = None) -> bytes:
+    result = x.to_bytes((x.bit_length() + 7) // 8, "big")
+    result = (
```
aorumbayev, line 86: should we prevent cases when pad_to is > than result ?

algopy_testing/src/algopy_testing/arc4.py
```diff
@@ -27,6 +32,231 @@ def wrapped(*args: _P.args, **kwargs: _P.kwargs) -> _R:
...
+
+        self.__max_bytes_len = self.__bit_size // BITS_IN_BYTE
+        bytes_value = int_to_bytes(value, self.__max_bytes_len)
+        self.__value = as_bytes(bytes_value, max_size=self.__max_bytes_len)
+
+    @classmethod
+    def from_bytes(cls, value: algopy.Bytes | bytes, /) -> typing.Self:
+        """Construct an instance from the underlying bytes (no validation)"""
+        value = as_bytes(value)
+        result = cls(int.from_bytes(value))
```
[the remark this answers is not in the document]
boblat: `from_bytes` is a class method and I think it doesn't have access to the `__max_bytes_len` attribute until after an instance of the class is created. 

The check for the input bytes length is implemented to emulate the ABI encoding behaviour which raises a logic error if the input bytes exceeds the maximum length.

`__init__` of `UIntN` and `BigUIntN` classes  check if this integer value exceeds the maximum value that can be represented by the number of bits specified by the generic type of the class.  However, they cannot check if the `bytes` value given to `from_bytes` method exceeds the maximum length of bytes supported by the class under construction.

algopy_testing/src/algopy_testing/utils.py
boblat: I don't think we should. For example, `int_to_bytes(1, 8)` should return `b'\x00\x00\x00\x00\x00\x00\x00\x01'`.

algopy_testing/src/algopy_testing/arc4.py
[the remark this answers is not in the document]
boblat: I found that I was not testing AVM implementation of `from_bytes`. It actually did not raise a logic error if the input bytes length did not match the data types' length. It returned `None` which was the same behaviour as `from_log`. I've removed the length check in the stub implementation in this commit d86cb31c2846c5d43382c6b2a0d4cf08c603ef80.

algopy_testing/tests/arc4/test_uintn.py
```diff
@@ -0,0 +1,238 @@
...
+        (int_to_bytes(2**16, 4), ARC4_RETURN_PREFIX, 2**16),
+        (int_to_bytes(2**32 - 1, 4), ARC4_RETURN_PREFIX, 2**32 - 1),
+        (int_to_bytes(255, 4), b"", None),
+        (int_to_bytes(255, 4), b"\xff\x00\x01\x02", None),
+    ],
+)
+def test_uintn_from_log(
+    get_avm_result: AVMInvoker, value: bytes, prefix: bytes, expected: int | None
+) -> None:
+    if expected is None:
```
daniel-makerx, line 198: Seperate this into two tests? one for the error scenarios and one for the happy path?

boblat: They have been split into separate test cases as discussed in this commit 1251f4eb4e16e680d4315b3f30e50b2d98640e0c.

aorumbayev:
@daniel-makerx when trying out the stubs generator script (algopy_testing/scripts/check_stubs_cov) with @boblat we found out that `_get_impl_coverage` has an edge case on the UIntN when performing getattr on the algopy_testing.arc4 module. However the collections.abc also contains an abstraction of the same name and getattr in such cases always returns the wrong abstraction (from abc module instead of our implementation in arc4.py). 

Do you know if there are any clean workaround in such cases? Otherwise we can use a list of exceptions for the time being so it uses direct imports instead of getattr (specifically when it encounetrs a module named arc4)

daniel-makerx:
> @daniel-makerx when trying out the stubs generator script (algopy_testing/scripts/check_stubs_cov) with @boblat we found out that `_get_impl_coverage` has an edge case on the UIntN when performing getattr on the algopy_testing.arc4 module. However the collections.abc also contains an abstraction of the same name and getattr in such cases always returns the wrong abstraction (from abc module instead of our implementation in arc4.py).
> 
> Do you know if there are any clean workaround in such cases? Otherwise we can use a list of exceptions for the time being so it uses direct imports instead of getattr (specifically when it encounetrs a module named arc4)

This is a bit of a quirk of how specialized generics are represented, I'll add a fix

PR #237, opened 2024-05-30
fix: fixes issues when reading and writing dynamic ARC4 types in an array

daniel-makerx:
When reading a dynamic ARC4 tuple/struct from an array the underlying element was being incorrectly assumed to be another array, this resulted in incorrect offsets being used and often a logic error when extracting the value. This PR addresses that issue


achidlow (approved):

src/puya/ir/optimize/intrinsic_simplification.py
```diff
@@ -227,7 +227,10 @@ def _try_fold_intrinsic(
                 binary_array = [
                     x for xs in [bin(bb)[2:].zfill(8) for bb in byte_const.value] for x in xs
                 ]
-                binary_array[index] = "1" if value else "0"
+                try:
+                    binary_array[index] = "1" if value else "0"
+                except IndexError:
+                    return None  # would fail at runtime
```
achidlow, line 233: this would be a good use case for a poison-pill or similar

PR #238, opened 2024-06-01
chore: refining hatch venvs on algorand-python-testing

aorumbayev:
## Proposed Changes

- Minor tweaks to make default env on algopy_testing work with examples folder 
- The examples folder is to be used as part of the CI CD only


aorumbayev:
@daniel-makerx tried to follow the suggestions we discussed but the algorand-python-testing still not reloading the content live - require pruning examples venv. Let me know what you think could be the root cause, i'll dig into hatch docs a bit more as well

daniel-makerx:
> @daniel-makerx tried to follow the suggestions we discussed but the algorand-python-testing still not reloading the content live - require pruning examples venv. Let me know what you think could be the root cause, i'll dig into hatch docs a bit more as well


If you are still running using the examples venv, then it will be using the dependencies as defined [here](https://github.com/algorandfoundation/puya/pull/238/files#diff-ecb4eb66a4622e1278e2bd9a5808291983ab4b41e69c1707cde290a10a21234dR129), which won't reflect changes until a hatch prune is done.

However my thinking is while actively developing the library just use the default venv (with the now editable dependencies), and just use the examples venv (with non editable dependencies) as part of CI/CD

aorumbayev:
@daniel-makerx given we went over this over a call and small # of changes ill merge this to state branch and will await for final review comments there 👍

PR #241, opened 2024-06-06
feat: stub implementation of arc4 UFixedNxM, BigUFixedNxM, String and Bool

boblat:
## Proposed Changes

Implement stubs for
- UFixedNxM 
- BigUFixedNxM
- String
- Bool


algopy_testing/src/algopy_testing/arc4.py
```diff
@@ -59,6 +58,55 @@ def from_log(cls, log: algopy.Bytes, /) -> typing.Self:
...
 
 
+class String(_ABIEncoded):
+    """An ARC4 sequence of bytes containing a UTF8 string"""
+
+    _value: bytes
+
+    def __init__(self, value: algopy.String | str = "", /) -> None:
+        value = as_string(value)
+        bytes_value = value.encode("utf-8")
```
daniel-makerx, line 68: To preserve semantic compatibility between these stubs and the compiled behaviour this should only encode value to bytes for `str` type, for the `algopy.String` type it should use the underlying bytes of that type, this probably means changing the `algopy.String` implementation too to support this. Any other type should be a TypeError

algopy_testing/src/algopy_testing/arc4.py
```diff
@@ -59,6 +58,55 @@ def from_log(cls, log: algopy.Bytes, /) -> typing.Self:
...
+        self._value = len(bytes_value).to_bytes(_ABI_LENGTH_SIZE) + bytes_value
+
+    @property
+    def native(self) -> algopy.String:
+        """Return the String representation of the UTF8 string after ARC4 decoding"""
+        return algopy.String(self._value[_ABI_LENGTH_SIZE:].decode("utf-8"))
```
daniel-makerx, line 74: Again to preserve semantic compatibility here this should avoid doing the decode operation, and instead construct an `algopy.String` using just the bytes

algopy_testing/src/algopy_testing/arc4.py
```diff
@@ -223,6 +271,123 @@ def __bool__(self) -> bool:
         return bool(self.native)
 
 
+_TDecimalPlaces = typing.TypeVar("_TDecimalPlaces", bound=int)
+_MAX_M_SIZE = 160
+
+
+class _UFixedNxMMeta(type(_ABIEncoded), typing.Generic[_TBitSize, _TDecimalPlaces]):  # type: ignore  # noqa: PGH003
```
daniel-makerx, line 278: Can this ignore the specific type violations using `# type: ignore[RULE1, RULE2, ...]` instead of PGH003?

algopy_testing/src/algopy_testing/arc4.py
```diff
@@ -223,6 +271,123 @@ def __bool__(self) -> bool:
...
+    __concrete__: dict[tuple[type[_TBitSize], type[_TDecimalPlaces]], type] = {}  # noqa: RUF012
```
daniel-makerx, line 279: wrapping the type annotation in `typing.ClassVar` should remove the need to ignore RUF012

algopy_testing/src/algopy_testing/arc4.py
boblat: It seems unnecessary and has been removed in this commit 2cacca23751e39919ec2ec76e26d927599c05f16.

algopy_testing/src/algopy_testing/arc4.py
boblat: It has been fixed in this commit daab1b78c560ff747c4569d4f44321ad180e70f7.

algopy_testing/src/algopy_testing/arc4.py
boblat: It has been fixed in this commit daab1b78c560ff747c4569d4f44321ad180e70f7.

algopy_testing/src/algopy_testing/arc4.py
boblat: It has been fixed in this commit 682a52594452b917366e4f3557ec65e49fb58dce.

algopy_testing/src/algopy_testing/primitives/string.py
```diff
@@ -12,10 +12,11 @@ class String:
     indexing and length operations are not supported. Use .bytes.length for byte length.
     """
 
-    value: str
+    _value: bytes  # underlying 'bytes' value representing the String
 
     def __init__(self, value: str = "") -> None:
-        self.value = as_string(value)
+        value = as_string(value)
```
daniel-makerx, line 18: Unfortunately can't use as_string here as this should strictly only be the `str` type, `as_string` would also allow `algopy.String` and `algopy.arc4.String`

algopy_testing/src/algopy_testing/arc4.py
```diff
@@ -50,9 +49,61 @@ def from_log(cls, log: algopy.Bytes, /) -> typing.Self:
...
+class String(_ABIEncoded):
+    """An ARC4 sequence of bytes containing a UTF8 string"""
+
+    _value: bytes
+
+    def __init__(self, value: algopy.String | str = "", /) -> None:
+        bytes_value = (
+            value.bytes.value
+            if isinstance(value, algopy.String)
+            else as_string(value).encode("utf-8")
```
daniel-makerx, line 61: Unfortunately can't use as_string here either as it allows broader types than the stubs indicate.

Can use a `match` to handle `str` and `algopy.String` and then the default case can just be a type error

algopy_testing/src/algopy_testing/primitives/string.py
boblat: It has been fixed in this commit 9e8f1dfaa2b440b9ef3241ad3290abf7df87bce9.

algopy_testing/src/algopy_testing/arc4.py
boblat: It has been fixed in this commit 9e8f1dfaa2b440b9ef3241ad3290abf7df87bce9.

PR #250, opened 2024-06-16
feat: defaults for transaction and globals; gtxn and itxn group refinements; tests for non arc4 contracts

aorumbayev:
## Proposed Changes

- Sets default values to Global object, defaults to base transaction fields (txid and etc)
- Auto adding instantiated contracts inside context into context state, auto adding contract calls as application call transactions to active context transaction group 
- Auto adding inner transactions to context, allowing user to quickly inspect inner transaction groups after execution of the method targeted for testing


algopy_testing/src/algopy_testing/context.py
```diff
@@ -168,54 +232,103 @@ def update_asset(self, asset_id: int, **asset_fields: Unpack[AssetFields]) -> No
...
 
         Args:
-            itxn (BaseInnerTransaction): The inner transaction to be added.
+            contract (algopy.Contract | algopy.ARC4Contract): The contract to add.
+        """
+        self.contracts.append(contract)
+
+    def append_inner_transaction_group(
+        self,
+        itxn: typing.Sequence[
```
daniel-makerx, line 281: use `collections.abc.Sequence` instead

algopy_testing/src/algopy_testing/context.py
```diff
@@ -168,54 +232,103 @@ def update_asset(self, asset_id: int, **asset_fields: Unpack[AssetFields]) -> No
...
+
+        Args:
+            itxn (Sequence[InnerTransactionResult | PaymentInnerTransaction |
+            KeyRegistrationInnerTransaction | AssetConfigInnerTransaction
+            | AssetTransferInnerTransaction | AssetFreezeInnerTransaction
+            | ApplicationCallInnerTransaction]): The group of inner transactions to append.
+        """
+        import algopy.itxn
+
+        self.inner_transaction_groups.append(cast(list[algopy.itxn.InnerTransactionResult], itxn))
```
daniel-makerx, line 302: rather than casting, could make a union type at the module level `InnerTransactionResultType = algopy.itxn.InnerTransactionResult | algopy.itxn.PaymentTransactionResult | ...`

Then make the list on the context this type
selft.inner_transaction_groups `list[Sequence[InnerTransactionResultType]]`

and lastly make the itxn arg `Sequence[InnerTransactionResultType]`

algopy_testing/src/algopy_testing/gtxn.py
```diff
@@ -2,133 +2,223 @@
...
+                "type_bytes": txn_type_to_bytes(int(algopy.TransactionType.AssetConfig)),
+                **kwargs,
+            }
+        )
+
+    def __getattr__(self, name: str) -> typing.Any:  # noqa: ANN401
+        if name in AssetConfigFields.__annotations__:
+            return self._fields.get(name)
+
+        raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")
```
daniel-makerx, line 176: use `type(self)` instead of `self.__class__`

algopy_testing/src/algopy_testing/itxn.py
```diff
@@ -1,284 +1,349 @@
...
-    fee: algopy.UInt64
-    note: algopy.String
-    rekey_to: algopy.Account
+    def _parse_covariant_types(
+        self,
+    ) -> None:
+        import algopy
 
+        for name, value in self.fields.items():
+            if isinstance(value, int):
```
daniel-makerx, line 248: Need to handle `str` here too

algopy_testing/src/algopy_testing/models/contract.py
```diff
@@ -1,2 +1,78 @@
...
+
+class ARC4Contract(Contract):
+    @final
+    def approval_program(self) -> algopy.UInt64 | bool:
+        # Implement the logic for the approval program
+        from algopy import UInt64
+
+        return UInt64(1)  # Example return value
+
+    def clear_state_program(self) -> algopy.UInt64 | bool:
```
daniel-makerx, line 74: The default clear state should return `True`

algopy_testing/src/algopy_testing/models/contract.py
```diff
@@ -1,2 +1,78 @@
...
+    ):
+        cls._name = name or cls.__name__
+        cls._scratch_slots = scratch_slots
+        cls._state_totals = state_totals
+
+    def approval_program(self) -> algopy.UInt64 | bool:
+        # Implement the logic for the approval program
+        from algopy import UInt64
+
+        return UInt64(1)  # Example return value
```
daniel-makerx, line 57: These two methods on Contract should be either abstract (but that might be tricky with the custom metaclass) or just raise `NotImplementedError` as the puya compiler does not provide defaults for these methods

algopy_testing/src/algopy_testing/models/contract.py
```diff
@@ -1,2 +1,78 @@
...
+        from algopy import UInt64
+
+        return UInt64(0)  # Example return value
+
+
+class ARC4Contract(Contract):
+    @final
+    def approval_program(self) -> algopy.UInt64 | bool:
+        # Implement the logic for the approval program
+        from algopy import UInt64
```
daniel-makerx, line 70: Ideally to match Puya this should actually perform routing based on the current Txn args, app_id and on completion action, however that is a fair bit of effort for little value. 

So might be better to raise a NotImplementedError here instead directing the user to call the ARC4 methods directly, thoughts?

algopy_testing/src/algopy_testing/models/itxn.py
```diff
@@ -13,12 +12,12 @@ def __getattr__(self, name: str) -> Callable[[], typing.Any]:
...
                 "the context manager."
             )
-        if not context.inner_transactions:
+        if not context.inner_transaction_groups:
             raise ValueError(
                 "No inner transaction found in the context! Use `with algopy_testing_context()` "
                 "to access the context manager."
             )
-        itxn = context.inner_transactions[-1]
+        itxn = context.inner_transaction_groups[-1]
```
daniel-makerx, line 20: This should get the last transaction of the last group e.g. `context.inner_tranasctions_groups[-1][-1]` but with appropriate checks

daniel-makerx: Additionally we should also implement GITxn in a similar fashion, in GITxn the first argument for each method is the index of the transaction within the last group

algopy_testing/src/algopy_testing/models/contract.py
aorumbayev: @daniel-makerx agree, adding notimplementederror here by default, in most cases when dealing with ARC4Contract a user would prefer direct method call anyway

algopy_testing/src/algopy_testing/context.py
aorumbayev: @daniel-makerx i ended up only addressing this partially, the reason i went for the cast in the first place is because the algopy.itxn.InnerTransactionResult is the only non protocol class in the itxn module. The rest are all non runtime-checkable protocols, hence using a tuple works well but the instance user receives won't work with methods like isinstance and you will have to do some sort of casting regardless if you get a union type in return. I still apply the suggestion on InnerTransactionResultType but only on the append method to make it more concise, the inner_transaction_group is using Sequence as suggested however i still do the casting and returning object type from this method is algopy.itxn.InnerTransactionResult, for type checking user can refer to the type enum and InnerTransactionResult conveniently encapsulates all fields from all transactions. 

Let me know what you think on this @daniel-makerx

PR #255, opened 2024-06-21
refactor: adding submit_txns, template vars, ScratchSlot, BytesBacked, baremethod, context improvements

aorumbayev:
## Proposed Changes

- Improving context manager, removing need for group_index
- GITxn,  submit_txns, template vars, ScratchSlot, BytesBacked, baremethod, arc4_signature and various tweaks
- Adding 'real' contract from ecosystem as a base for another example for tests https://github.com/giuliop/whitelist-example-algoplonk


aorumbayev:
@daniel-makerx would appreciate if you can take a look, adds a range of various tweaks. Will add unit tests and patch the zk verifier example contract tests tomorrow

algopy_testing/src/algopy_testing/context.py
```diff
@@ -74,33 +85,106 @@ class ContractContextData:
...
+class InnerTxnLoader:
+    """
+    Stores inner transaction references.
+    """
+
+    def __init__(self, inner_txn_group: Sequence[InnerTransactionResultType]):
+        self.inner_txn_group = inner_txn_group
+
+    def _get_inner_transaction(self, txn_type: type[T]) -> T:
+        txn = next(
```
daniel-makerx, line 98: This behaviour could be misleading, would be safer to just get the last transaction and error if it isn't the required type?

algopy_testing/src/algopy_testing/context.py
```diff
@@ -74,33 +85,106 @@ class ContractContextData:
...
+                if isinstance(txn, txn_type)  # type: ignore[arg-type, unused-ignore]
+            ),
+            None,
+        )
+        if txn is None:
+            raise ValueError(f"{txn_type.__name__} not found in inner transaction group!")
+
+        return txn
+
+    def payment(self) -> algopy.itxn.PaymentInnerTransaction:
```
daniel-makerx, line 111: This could be properties?

algopy_testing/src/algopy_testing/op.py
```diff
@@ -568,6 +568,47 @@ def json_object(a: Bytes | bytes, b: Bytes | bytes, /) -> Bytes:
...
+
+        context = get_test_context()
+        slot_content = context._scratch_space[a]
+        match slot_content:
+            case Bytes():
+                return slot_content
+            case bytes():
+                return Bytes(slot_content)
+            case UInt64() | int():
+                return itob(slot_content)
```
daniel-makerx, line 584: If the stored value is the wrong stack type it should be an error, same with load_uint64

algopy_testing/src/algopy_testing/models/gitxn.py
```diff
@@ -0,0 +1,42 @@
...
+        if index >= len(itxn_group):
+            raise IndexError("Transaction index out of range")
+        itxn = itxn_group[index]
+        value = getattr(itxn, self._map_fields(name))
+        if value is None:
+            raise ValueError(f"'{name}' is not defined for {type(itxn).__name__}")
+        return value
+
+
+ITxn = _GITxn()
```
daniel-makerx, line 42: This should be `GITxn`

algopy_testing/src/algopy_testing/itxn.py
```diff
@@ -233,26 +411,91 @@ def __init__(
...
-            if isinstance(value, str):
-                self.fields[name] = algopy.Bytes(value.encode("utf-8"))
+            if name in relevant_fields:
+                if isinstance(value, int):
+                    self.fields[name] = algopy.UInt64(value)
+                elif isinstance(value, bytes):
+                    self.fields[name] = algopy.Bytes(value)
+                elif isinstance(value, str):
+                    self.fields[name] = algopy.Bytes(value.encode("utf-8"))
+                elif isinstance(value, tuple) and all(isinstance(v, bytes) for v in value):
```
daniel-makerx, line 480: Should check for tuple first and then iterate each item then convert as tuples can contain a mixture of python and algopy types

algopy_testing/examples/auction/test_contract.py
```diff
@@ -24,9 +24,8 @@ def test_opt_into_asset(context: AlgopyTestContext) -> None:
     contract.opt_into_asset(asset)
 
     # Assert
-    assert len(context.get_transaction_group()) == 1
     assert contract.asa.id == asset.id
-    inner_txn = context.get_last_submitted_inner_transaction()
+    inner_txn = context.get_last_submitted_itxn_loader().asset_transfer()
```
robdmoore, line 28: why `_loader`?

algopy_testing/examples/auction/test_contract.py
```diff
@@ -24,9 +24,8 @@ def test_opt_into_asset(context: AlgopyTestContext) -> None:
...
     assert inner_txn.type == algopy.TransactionType.AssetTransfer
```
robdmoore, line 29: This one shouldn't be needed right?

algopy_testing/src/algopy_testing/context.py
robdmoore: +1

algopy_testing/src/algopy_testing/context.py
```diff
@@ -605,51 +693,92 @@ def any_app_call_txn(  # noqa: PLR0913
...
+        import algopy.gtxn
+
+        new_txn = algopy.gtxn.KeyRegistrationTransaction(NULL_GTXN_GROUP_INDEX)
+
+        for key, value in kwargs.items():
+            setattr(new_txn, key, value)
+
+        return new_txn
+
+    def any_afrz_transaction(
```
robdmoore, line 773: Should we use `asset_freeze` rather than `afrz`, I assume we use full names in algorand python?

robdmoore: Same for the others, e.g. acfg etc.

algopy_testing/examples/auction/test_contract.py
aorumbayev: @robdmoore Regarding this, if I do remove loader wouldn't that be a bit misleading, the intent was to make it obvious that its not gonna return the txn but something that gives you ability to access them.

algopy_testing/examples/auction/test_contract.py
aorumbayev: @robdmoore you mean i shouldn't explicitly know that its an asset transfer and instead to get_last_submitted_itxn_loader().transaction()?

algopy_testing/src/algopy_testing/context.py
aorumbayev: @robdmoore the terms are not gonna be well knows for newcomers I agree, on the other hands they are fairly well understood by existing devs. The only downside with full names is more verbosity in invoking these methods. But other than that can rename them to full names

robdmoore: We want puya to be approachable for new devs. Verbosity doesn't matter vs the clarity in my opinion

algopy_testing/examples/auction/test_contract.py
robdmoore: The intellisense guides us on what the properties that are available are though? Loader is a weird concept that we would have to explain, easier and terser to not need to?

robdmoore: Or maybe we change it to:

```suggested
    inner_txn = context.last_submitted_itxn().get_asset_transfer()
```

robdmoore: Side-note: can the `last_submitted_itxn` bit be a property rather than a method? Or is that weird Python?

i.e.

```suggestion
    inner_txn = context.last_submitted_itxn.get_asset_transfer()
```

robdmoore: Also, out of curiosity, what's the syntax when getting an itxn by index? Do we have this same loader abstraction?

i.e. something like

```python
    inner_txn = context.submitted_itxns[2].get_asset_transfer()
```

algopy_testing/examples/proof_of_attendance/contract.py
```diff
@@ -0,0 +1,80 @@
+import algopy
+
+
+class ProofOfAttendance(algopy.ARC4Contract):
+    def __init__(self) -> None:
+        self.max_attendees = algopy.UInt64(30)
+        self.asset_url = algopy.String("ipfs://QmW5vERkgeJJtSY1YQdcWU6gsHCZCyLFtM1oT9uyy2WGm8")
+        self.total_attendees = algopy.UInt64(0)
+        self.initialized = algopy.UInt64(0)
```
daniel-makerx, line 9: can use a bool here e.g. `False`

PR #259, opened 2024-06-25
`algorand-python-testing` minor refinements; fixing links in algopy testing docs; simplifying auction tests

aorumbayev:


aorumbayev:
@daniel-makerx this is just a followup to today's 0.2.0, will release under 0.2.1, patches minor extra items in the documentation for the unit testing package as well as simplifying the tests and patching an edge case for encoding return abi type when assembling a method signature

aorumbayev:
@daniel-makerx i think introducing hatch in stubs broke some configuration on root mypy, im not sure what the root cause though, and why it didn't fail on initial merge on master branch. Could it be a case of it reusing an older cache ? 

Can't seem to replicate it locally though, really strange bug, the changes in this PR aren't overlapping with the parts that are failing.

daniel-makerx (approved):

PR #262, opened 2024-07-04
feature: Refactor awst node for switch blocks to support multiple clauses per block

tristanmenzel:
## Proposed Changes

JavaScript/TypeScript (and several other languages) allow multiple cases per case block. They also support case fall through. The latter is [discouraged](https://eslint.org/docs/latest/rules/no-fallthrough) as it can often be undesired behaviour leading to a bug. The former however is highly desirable as it leads to terser code.

eg.

```ts
switch(someValue) {
  case 'value1':
  case 'value2':
     return doSomething()
  case 'value3':
     return doSomethingElse()
  default:
     return defaultValue()
```

This PR replaces the `Mapping[Expression, Block]` of the awst `Switch` node with a sequence of `SwitchCaseBlock` nodes which support several clauses to be listed per block in a `Sequence[Expression]`. 

At the IR build level these are mapped out to individual cases which `goto` the same `BasicBlock`. 

A small tweak was required for the IR `Switch` model to return only distinct blocks from its `targets` property.

---

The changes here result in no changes to the output for existing python contracts and I have done some basic testing of multiple clauses by manipulating the awst in the awst_build step but otherwise it's hard to get additional test coverage without hooking up another language front end.

### DIFF
--- a/src/puya/awst/function_traverser.py
+++ b/src/puya/awst/function_traverser.py
@@ -199,9 +199,10 @@ class FunctionTraverser(
 
     def visit_switch(self, statement: awst_nodes.Switch) -> None:
         statement.value.accept(self)
-        for case, block in statement.cases.items():
-            case.accept(self)
-            block.accept(self)
+        for case_block in statement.cases:
+            for clause in case_block.clauses:
+                clause.accept(self)
+            case_block.block.accept(self)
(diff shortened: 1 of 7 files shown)


tristanmenzel:
Setting to draft until we decide how to handle case fall throughs

PR #266, opened 2024-07-09
refactor: Add labeled blocks with goto statement

tristanmenzel:
## Proposed Changes

- Add label property to blocks so they can be 'referenced' by name
- Add goto node for unconditional navigation to a labelled block
-

### DIFF
--- a/src/puya/awst/function_traverser.py
+++ b/src/puya/awst/function_traverser.py
@@ -13,6 +13,9 @@ class FunctionTraverser(
     def visit_copy(self, expr: awst_nodes.Copy) -> None:
         expr.value.accept(self)
 
+    def visit_goto(self, statement: awst_nodes.Goto) -> None:
+        pass
+
     def visit_assignment_expression(self, expr: awst_nodes.AssignmentExpression) -> None:
         expr.target.accept(self)
         expr.value.accept(self)
--- a/src/puya/ir/builder/blocks.py
+++ b/src/puya/ir/builder/blocks.py
@@ -41,6 +41,8 @@ class BlocksBuilder:
         self.ssa = BraunSSA(blocks, parameters, self.active_block)
         self.ssa.seal_block(self.active_block)
 
+        self._labelled_blocks = dict[str, BasicBlock]()
+
     @property
     def blocks(self) -> Sequence[BasicBlock]:
         return self._blocks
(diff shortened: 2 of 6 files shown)


src/puya/ir/builder/blocks.py
```diff
@@ -164,3 +166,30 @@ def loop_continue(self, source_location: SourceLocation) -> None:
...
+        for block in self._labelled_blocks.values():
+            self.ssa.seal_block(block)
+
+    def goto_label(self, label: str, source_location: SourceLocation) -> None:
+        try:
+            target = self._labelled_blocks[label]
+            self.goto(target, source_location)
+        except KeyError:
+            target = self._labelled_blocks[label] = BasicBlock(
+                source_location=source_location, comment="Temp"
```
tristanmenzel, line 193: Can't use the correct source_location or comment here as it's not known at the time. We could 'replace' the block once it is known?

src/puya/ir/builder/blocks.py
```diff
@@ -164,3 +166,30 @@ def loop_continue(self, source_location: SourceLocation) -> None:
...
+    def make_labelled_block(
+        self, label: str, source_location: SourceLocation, description: str | None
+    ) -> BasicBlock:
+        try:
+            new_block = self._labelled_blocks[label]
+        except KeyError:
+            new_block = self._labelled_blocks[label] = BasicBlock(
+                source_location=source_location, comment=description
+            )
+        self.goto(new_block)
```
daniel-makerx, line 179: could add an else block here and set source_location and comment on the new_block to address the issue you mentioned below?

src/puya/awst/function_traverser.py
```diff
@@ -1,5 +1,6 @@
 import puya.awst.visitors
 from puya.awst import nodes as awst_nodes
+from puya.awst.visitors import T
```
daniel-makerx, line 3: Shouldn't need this?

src/puya/ir/builder/blocks.py
tristanmenzel: oh for some reason I thought ir nodes were immutable

src/puya/awst/function_traverser.py
tristanmenzel: Just a remnant of pycharm auto creating the missing abstract member

tristanmenzel:
closing in favour of #279

PR #268, opened 2024-07-12
feature: Include postfix unary operations for uint64 and biguint

tristanmenzel:
## Proposed Changes

- Adds support for `i++` and `i--` expressions on uint64 and biguint types

### DIFF
--- a/src/puya/ir/builder/main.py
+++ b/src/puya/ir/builder/main.py
@@ -238,6 +238,52 @@ class FunctionIRBuilder(
         else:
             return ValueTuple(expr.source_location, list(result))
 
+    def visit_biguint_postfix_unary_operation(
+        self, expr: awst_nodes.BigUIntPostfixUnaryOperation
+    ) -> TExpression:
+        target_value = self.visit_and_materialise_single(expr.target)
+        rhs = BigUIntConstant(value=1, source_location=expr.source_location)
+        match expr.op:
+            case awst_nodes.BigUIntPostfixUnaryOperator.increment:
+                binary_op = awst_nodes.BigUIntBinaryOperator.add
+            case awst_nodes.BigUIntPostfixUnaryOperator.decrement:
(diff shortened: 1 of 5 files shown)


src/puya/ir/builder/main.py
```diff
@@ -165,6 +165,98 @@ def visit_assignment_expression(self, expr: awst_nodes.AssignmentExpression) ->
         else:
             return ValueTuple(expr.source_location, list(result))
 
+    def visit_biguint_postfix_unary_operation(
```
tristanmenzel, line 168: There's a lot of repeated code here but it's different enough that it might not be worth refactoring

src/puya/ir/builder/main.py
```diff
@@ -238,6 +238,98 @@ def visit_assignment_expression(self, expr: awst_nodes.AssignmentExpression) ->
...
+        self, expr: awst_nodes.BigUIntPostfixUnaryOperation
+    ) -> TExpression:
+        target_value = self.visit_and_materialise_single(expr.target)
+        rhs = BigUIntConstant(value=1, source_location=expr.source_location)
+        match expr.op:
```
tristanmenzel, line 246: is it better to do this with a dictionary wrapped in `try ... except KeyError` ? I guess we lose compile time safety over the exhaustive check

achidlow: I prefer the way you've got it, given it's located in an entirely different layer (IR) to where it's defined (AWST)

src/puya/ir/builder/main.py
achidlow: Agreed, not worth it

achidlow:
Looking good, but can we roll the prefix operators up with the existing unary operators in the AWST layer? The existing unary ops are prefix as well. Having the distinction between postfix and prefix encoded in the types is good though, I like that you've done that.

achidlow (approved):

PR #275, opened 2024-07-19
fix a regression where `algopy.op` functions that accepted multiple literal types would fail to compile with all except one type

### LINKED ISSUE #274: regression: `algopy.op` functions that take multiple literal types no longer compile for all accepted types
e.g.

```python3
op.Scratch.store(7, 2)
```

gives the error:
```
can't covert literal to algopy.Bytes
            op.Scratch.store(7, 2)
```

A current work-around is to ensure all such values are wrapper in their corresponding `algopy` type e.g.:

`op.Scratch.store(7, UInt64(2))`

achidlow:
Fixes #274 

This was due to the changed behaviour to log-and-continue with a dummy value rather than throw on failed conversions


src/puya/awst_build/eb/interface.py
```diff
@@ -117,6 +117,15 @@ def resolve_literal(self, converter: TypeBuilder) -> InstanceBuilder:
         or if the literals are not expected to be homogenous (e.g. in the general case of a tuple
         with literals)"""
 
+    @abc.abstractmethod
+    def try_resolve_literal(self, converter: TypeBuilder) -> InstanceBuilder | None:
+        """Similar to resolve_literal, but in the case where a conversion is possible but the
+        literal values are the wrong type, don't produce any errors and return None instead.
+
+        If not conversion is necessary (ie no literals or it's not meaningful to convert literals
```
daniel-makerx, line 125:
```suggestion
        If no conversion is necessary (ie no literals or it's not meaningful to convert literals
```

achidlow: Thanks

daniel-makerx (approved):

PR #276, opened 2024-07-22
Fix assemble when last op is a branch op

daniel-makerx:


src/puya/ussemble/output.py
```diff
@@ -89,6 +92,10 @@ def _get_offset_from_end(self, label: models.Label) -> int:
         try:
             return self._op_pc[label_index] - self._op_pc[self._op_index + 1]
         except KeyError:  # during first pass offsets will not be present
+            if not self._is_first_pass:
```
achidlow, line 95: will this always error if it's the first pass? if so, maybe we just return that straight away

daniel-makerx: yes it will on the first pass, due to the lookup of `self._op_index + 1`. Have changed it to return early instead

achidlow (approved):

PR #277, opened 2024-07-23
feat: include ARC-22 readonly flag in ARC-32 output

daniel-makerx:


src/puya/client_gen.py
```diff
@@ -94,7 +94,12 @@ def parse_app_spec_methods(app_spec_json: str) -> tuple[str, Sequence[ARC4Method
                     source_location=None,
                     name=arc4_method.signature.name,
                     create=create_option,
-                    readonly=bool(method_hints.get("read_only")),
+                    readonly=bool(
```
achidlow, line 97: As discussed, let's take the or of the two values, otherwise this PR looks good to go.

daniel-makerx: Done

achidlow (approved):

PR #279, opened 2024-07-23
while-else and for-else loop support

achidlow:
Support Python for-else and while-else loop constructs, through the use of a new Goto node which works by targeting an AWST block which contains the given label.

This is also intended to be used by languages with switch/cases that have fall-through semantics, or for targeted loop breaking as per Javascript support.

Block building code has been refactored to be safer w.r.t. sealing automatically at an appropriate time.

Iteration code has been refactored for simplicity and also improved code generation.


src/puya/awst/nodes.py
```diff
@@ -2,7 +2,6 @@
 import enum
```
tristanmenzel, line 2: Is it worth dropping the break and continue nodes to avoid ambiguity for future front ends?

src/puya/ir/arc4_router.py
```diff
@@ -60,7 +60,7 @@
 def create_block(
     location: SourceLocation, description: str | None, *stmts: awst_nodes.Statement
```
tristanmenzel, line 61: should we propagate the rename of description => comment?

src/puya/ir/builder/blocks.py
```diff
@@ -41,18 +42,23 @@ def __init__(
...
     def add(self, op: Op) -> None:
         """Add an op"""
-        curr = self._get_current_for_addition(op.source_location)
-        if curr is not None:
+        curr = self.active_block
+        if curr.terminated:
+            logger.error(
+                "unreachable code",
+                # function kinda sucks as a fallback, but it'll do for now
+                location=op.source_location or self._default_source_location,
```
tristanmenzel, line 59: could maybe try `curr.terminator.source_location` before the function default as it would hopefully be closer to the actual problem

tristanmenzel (approved):
Couple of minor comments to consider, otherwise LGTM

src/puya/awst/nodes.py
achidlow: I think they're still valuable for straightforward loop construction, there's a bit of song and dance involved to replace their usage entirely with Goto.

To resolve the ambiguity, I've renamed the nodes now:
- `BreakStatement` -> `LoopExit`
- `ContinueStatement` -> `LoopContinue`
And also added a very short docstring noting that this applies to the innermost loop.

src/puya/awst_build/validation/labels.py
```diff
@@ -0,0 +1,42 @@
...
+    def validate(cls, module: awst_nodes.Module) -> None:
+        for module_statement in module.body:
+            if isinstance(module_statement, awst_nodes.Subroutine):
+                cls(module_statement)
+            elif isinstance(module_statement, awst_nodes.ContractFragment):
+                for method in module_statement.methods.values():
+                    cls(method)
+
+    def __init__(self, function: awst_nodes.Function) -> None:
+        self._function = function
```
daniel-makerx, line 21: Don't think this is needed?

src/puya/ir/arc4_router.py
achidlow: In this case, yep.

src/puya/ir/builder/blocks.py
achidlow: This was copied from existing code but that's a good point, I'll update this and the original code.

daniel-makerx (approved):

Issue #185, opened 2024-04-05
Support arc4.StaticArray empty constructors


tristanmenzel:
Given a static array of a size and type
```python
MyArray: typing.TypeAlias = StaticArray[Address, typing.Literal[2]]
```
There should be an empty constructor
```python
my_value = MyArray()
```
that is the equivalent of initializing the array with N items, initialized via their empty constructor
```python
MyArray() == MyArray(Address(), Address())
```
This should recurse for nested static arrays

PR #281, opened 2024-07-25
Allow staticarrays to be constructed with no arguments

tristanmenzel:
## Proposed Changes

- Addresses #185 
-
-


test_cases/static_arrays_no_args/contract.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+        (a, b, c) = Application(), Asset(), Account()
+
+        assert StaticArray[UInt32, typing.Literal[1]]() == StaticArray(UInt32())
+
+        assert StaticArray[DynamicArray[AddressArray], typing.Literal[1]]() == StaticArray(
+            DynamicArray[AddressArray]()
+        )
+        # TODO: The following should either work, or fail gracefully
+        # x = StaticArray[TestStruct, typing.Literal[4]]()
```
tristanmenzel, line 43: I presume it's possible to add a default ctor to the arc4.Structs but it maybe doesn't make much sense. Is it better to just error when attempting to construct a static array of struct without any values?

achidlow: It wouldn't jive with the datalcass vibe - not unless the class had default values for all fields (which we don't support yet)

PR #283, opened 2024-07-26
fix: handle comparison of boolean-evaluated UInt64 objects correctly 

### LINKED ISSUE #194: Comparisons with `algopy.UInt64` that have been converted to bool fail
The following assertion will fail, when it should pass:

```python3
assert bool(algopy.UInt64(1)) == bool(algopy.UInt64(2))
```

As will this one:
```python3
two = algopy.UInt64(2)
match bool(two):
    case True:
        pass
    case _:
        assert False
```

achidlow:
Fixes #194 

This fix necessarily increases code size - some optimisations were improved to account for this, see commit messages for details.


src/puya/awst_build/eb/uint64.py
```diff
@@ -92,16 +91,13 @@ def to_bytes(self, location: SourceLocation) -> Expression:
 
     @typing.override
     def bool_eval(self, location: SourceLocation, *, negate: bool = False) -> InstanceBuilder:
-        as_bool = ReinterpretCast(
-            expr=self.resolve(),
-            wtype=wtypes.bool_wtype,
-            source_location=self.resolve().source_location,
+        cmp_expr = NumericComparisonExpression(
```
tristanmenzel, line 94: Is there a scenario where you would get a python `bool` from an non boolean expression without calling `bool(<expr>)`? 

If not, then it might make sense to shift this logic into `BoolTypeBuilder.call`. This is what I'm doing in puya-ts currently but I would be keen to know if I'm missing something obvious.

achidlow: When it's a simple expression like this, I prefer to not couple the expression builders together. That plus `call` is effectively untyped, since it handles any call signature encountered in user code before verifying and extracting arguments.

tristanmenzel: I think you miss understood, I still use `bool_eval` to convert to a `<zero or non-zero>` type, but then use the `BooleanFunctionBuilder.call` method to do the additional `x.bool_eval() !== 0`.

daniel-makerx (approved):

src/puya/awst_build/eb/uint64.py
achidlow: The boolean type builder would still need to handle multiple types and construct different nodes then. That is, unless by "convert to zero or non-zero type" you mean converted everything to a uint64, in which case a simple utility function would also work rather than coupling builders together and going through an untyped function signature?

PR #287, opened 2024-08-06
SingleEvaluation node object equality

achidlow:
Make SingleEvaluation node work by structural equality rather than instance equality, to allow for (future) AWST serialization formats without references.

### DIFF
--- a/src/puya/awst/nodes.py
+++ b/src/puya/awst/nodes.py
@@ -803,24 +803,31 @@ class BoxValueExpression(Expression):
         return visitor.visit_box_value_expression(self)
 
 
-@attrs.frozen(init=False, eq=False, hash=False)  # use identity equality
+@attrs.frozen
 class SingleEvaluation(Expression):
-    """caveat emptor"""
+    """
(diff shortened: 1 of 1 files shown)


src/puya/awst/nodes.py
```diff
@@ -803,24 +803,31 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
...
-    """caveat emptor"""
+    """
+    This node wraps an underlying expression and effectively caches the result of that lowering,
+    such that regardless of how many times the SingleEvaluation object
+    (or any object comparing equal to it) appears in the AWST,
+    the underlying source expression will only be evaluated once.
+    """
 
     source: Expression
+    _id: int = attrs.field()
```
tristanmenzel, line 816: Does this need to be public for serialization to set it?

tristanmenzel (approved):

src/puya/awst/nodes.py
achidlow: nah, nothing is ever truly private in Python :P 
`cattrs` handles it fine.

PR #288, opened 2024-08-06
refactor: Change StateDelete node to be an expression

tristanmenzel:
## Proposed Changes

- Change StateDelete node to be an expression rather than statement as not all user languages have a `delete` statement with applicable use to this scenario

### DIFF
--- a/src/puya/awst/nodes.py
+++ b/src/puya/awst/nodes.py
@@ -1441,10 +1441,11 @@ class StateExists(Expression):
 
 
 @attrs.frozen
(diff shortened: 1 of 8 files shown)


achidlow:
Yep good call - this was originally just a `DeleteStatement` node to represent the Python delete statement, knowing other languages we cared about don't have such a statement but can make use of the underlying ops instead. However all that was used for was with state, so at some point it got turned into a `StateDelete` node but not changed into an `Expression`, which definitely makes more sense now.

PR #309, opened 2024-09-10
Fix source-location handling of user comments to be language independent

achidlow:
`SourceLocation` now has a `comment_lines` field to indicate how many lines to fetch as user comments when annotating output source files.

This provides a language independent way of handling source locations that can optionally include comments, whereas the current approach was to strip them out when necessary, assuming `#` to be the comment delimiter.

Also line numbers for code that has no source data are no longer output - this mostly means the embedded lib functions, for which the line numbers were not particularly useful to begin with.

Merging of source locations when operations are combined in optimisation has also been improved, to only merge locations when they are overlapping or adjacent, thus avoiding including unrelated code in annotations.

Specifically source locations with `self.` members has also been improved to use the location of the usage, rather than declaration, this previously resulted in extremely large and unrelated annotations in certain situations, see for examples the changes in this PR to `VotingRoundApp.approval.teal`.

Non-user facing changes:
- Improved SourceLocation validation, including making `end_line` non-optional (the existing behaviour of defaulting to `line` is retained, but is made explicit on the object).
- Fixed a bug with MIR output alignment caused by virtual ops, which made the column heading misleading.
- Fixed a bug with column locations when merging source locations.
- Fixed a bug where column locations were retained when including source comments.
- Fixed multiple places using user-comment locations when they should have been pointing at code only (e.g. IR labels, log output, ...)
- Validate rather than resolve SourceLocation paths, to prevent inconsistencies on Windows CI/CD and also to guard against accidental relative path inclusion especially when deserializing AWST


src/puyapy/awst_build/eb/contracts.py
```diff
@@ -158,3 +159,5 @@ def _builder_for_storage_access(
                     source_location=location,
                 ),
             )
+    result.source_location = location  # TODO: fixme
```
daniel-makerx, line 162: TODO or not TODO?

achidlow: Fixed, thanks!

daniel-makerx (approved):

PR #310, opened 2024-09-12
Fix template var error handling

achidlow:
`puyapy` fixes:
- handle zero values for TemplateVar
- boolean values are now also allowed as `True` or `False`
- incorrect template argument values will no longer result in a critical error, but a CLI usage error instead

`puya` changes:
- `PuyaOptions` now contains a parsed set of key-values for template variables

fixes for both:
- ensure exit code is set correctly when there are logged errors - this is an unreleased regression, so not marking as a fix for release notes purposes


src/puyapy/__main__.py
```diff
@@ -113,12 +116,11 @@ def main() -> None:
...
         metavar="VAR=VALUE",
-        action="append",
+        action=_ParseAndStoreTemplateVar,
+        nargs="+",
         help="Define template vars for use when assembling via --output-bytecode"
         " should be specified without the prefix (see --template-vars-prefix), e.g."
-        " -T=SOME_INT=1234"
-        " -T=SOME_BYTES=0x1A2B"
-        ' -T=SOME_STR=\\"hello\\"',
+        ' -T SOME_INT=1234 SOME_BYTES=0x1A2B SOME_BOOL=True SOME_STR=\\"hello\\"',
```
daniel-makerx, line 123: Should update docs too?

daniel-makerx (approved):

src/puyapy/__main__.py
achidlow: Docs look okay to me, there's only an example value there, and technically a bool is an int too in Python

PR #311, opened 2024-09-12
fix: optin oncomplete, assertions, method returns

iskysun96:
## Proposed Changes
- added `allow_actions=["OptIn"]` to the optin method decorator.
- Multiple assertions added for access control
- added returns on various methods 
- added `allow_actions=['DeleteApplication']` to the delete method. 

I didn't touch any of the output yet. Is there a workflow I can follow to regenerate the artifacts? Also are there any test files I need to edit?

### DIFF
--- a/examples/auction/contract.py
+++ b/examples/auction/contract.py
@@ -9,7 +9,6 @@ from algopy import (
     arc4,
     gtxn,
     itxn,
-    subroutine,
 )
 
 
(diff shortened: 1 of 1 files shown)


achidlow:
Thanks for the PR @iskysun96.

Changes look good, so just need to get the CI/CD passing.

First, you can run pre-commit on everything with `pre-commit run --all-files` to run ruff, black, and mypy. Looks like the only issue there is with formatting, so black should fix that up for you, and if you run it again all three should say "Passed".

The second thing is to update the outputs - we have a script that can do that, like so: `./scripts/compile_all_examples.py examples/auction`. Alternatively, you can run all tests via `pytest`, this will show failures where there are output differences vs what is committed, so if you commit those and run again they should all pass.

iskysun96:
Thanks @achidlow. I ran pre-commit and the script you shared to compile the new auction contract. 

Question. I ran pytest, and 700+ tests failed. Not sure what I am doing wrong.

achidlow:
I'm just guessing here, since I can't see the errors you got, but did you have localnet running?

Also you'll need to rebase your branch onto main before we can merge since there's been changes.

PR #312, opened 2024-09-13
Improve handling of `and`/`or` where the result is a type union

### LINKED ISSUE #145: Nested bool contexts give unexpected results
Currently, entering a "bool context" (for example, the target of an if condition, or explicitly converting something to `bool` via `bool(...)`) is not properly handled when nesting occurs.

```python3
from puyapy import Bytes, Contract, UInt64, op

class Oopsie(Contract):
        empty = Bytes()
        one = UInt64(1)
        return bool(op.bitlen(empty or one))

        return True
```

This will give the error message:

    error: Invalid argument type "bool" for argument "a" when calling puyapy.op.bitlen

This is incorrect, it's a type union not a `bool`. We allow type unions in a bool context explicitly, in this case it should warn that it's not supported (a further improvement is to actually support this particular case, which is feasible since the underlying op code supports the type union).

achidlow:
Expand handling of literal expressions to allow combining them with binary boolean operators, and improve error messaging when handling of type unions in nested bool contexts (fixes #145)

Also:
- allow dummy values to be constructed for literal-only pytypes
- single spot where "unexpected type" error message is defined, which also gives a special message in case of a union, just in case the function actually does support a union argument as is the case with some intrinsics etc
- prevent parsing function bodies when the return type is invalid, which can cause secondary errors which might be confusing and overshadow the root problem
- supress "expression result is ignored" warning when using `typing.assert_type`


src/puyapy/awst_build/pytypes.py
```diff
@@ -207,6 +207,35 @@ def wtype(self) -> typing.Never:
...
+        else:
+            result.append(t)
+    return unique(result)
+
+
+@typing.final
+@attrs.frozen
+class UnionType(PyType):
+    types: Sequence[PyType] = attrs.field(
+        converter=_flatten_nested_unions, validator=attrs.validators.min_len(1)
```
daniel-makerx, line 224: When would a union with only a single type occur? Or should this be `min_len(2)`?

src/puyapy/awst_build/eb/_expect.py
```diff
@@ -68,6 +67,15 @@ def defaulter(msg: str, location: SourceLocation) -> InstanceBuilder:  # noqa: A
     return defaulter
 
 
+def not_the_type_of(node: NodeBuilder, default: Callable[[str, SourceLocation], _T]) -> _T:
```
daniel-makerx, line 70: The name of this method sounds like it is checking the node is not a particular type (particular in the context of `expect.not_the_type_of`, but rather it is used to emit an error after that check has already been done. I think the function name should indicate that e.g. `log_incorrect_type` ?

src/puyapy/awst_build/pytypes.py
achidlow: It shouldn't no, and we probably do want a validation error rather than a code-error if it does, so yes, changed to 2.

src/puyapy/awst_build/eb/_expect.py
achidlow: Changed to `not_this_type` per discussion.

daniel-makerx (approved):

PR #316, opened 2024-09-19
refactor: Output a more useful error message when deserialization fails

tristanmenzel:
## Proposed Changes

- Improves error output when deserialization fails

### DIFF
--- a/src/puya/awst/serialize.py
+++ b/src/puya/awst/serialize.py
@@ -5,12 +5,16 @@ import typing
 from collections.abc import Mapping
 
 import cattrs
+from cattrs import ClassValidationError, IterableValidationError, transform_error
 from cattrs.preconf.json import make_converter
 from cattrs.strategies import configure_tagged_union, include_subclasses
 from immutabledict import immutabledict
 
+from puya import log
 from puya.awst import nodes, txn_fields, wtypes
 from puya.utils import StableSet
+
+logger = log.get_logger(__name__)
 
 
 def _unstructure_optional_enum_literal(value: object) -> object:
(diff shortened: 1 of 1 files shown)


src/puya/awst/serialize.py
```diff
@@ -86,4 +88,7 @@ def awst_to_json(awst: nodes.AWST) -> str:
 
 
 def awst_from_json(json: str) -> nodes.AWST:
-    return _get_converter().loads(json, nodes.AWST)  # type: ignore[type-abstract]
+    try:
+        return _get_converter().loads(json, nodes.AWST)  # type: ignore[type-abstract]
+    except (ClassValidationError, IterableValidationError, BaseException) as err:
```
achidlow, line 93: Never catch `BaseException` - it'll catch things like [KeyboardInterrupt](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt)

In general it's not even recommended to catch `Exception`, but sometimes that's warranted, and this might be one of those cases.

src/puya/awst/serialize.py
```diff
@@ -86,4 +88,7 @@ def awst_to_json(awst: nodes.AWST) -> str:
...
+        raise InternalError("Deserialization error: \n" + "\n".join(transform_error(err))) from err
```
achidlow, line 94: Since there's no source location, might as well make this a `ValueError`.

achidlow, line 94: But also, since this sort of error details would only really be useful for anyone working on the compiler(s), I'm wondering whether putting all this data into the exception message is the best experience in case anyone encounters it? It might be better to keep it succinct, and just log the details (ie results of `transform_error`) as debug messages.

src/puya/awst/serialize.py
tristanmenzel: Removed BaseException from clauses

src/puya/awst/serialize.py
tristanmenzel: changed to value error

tristanmenzel: Moved details of error into a `logger.debug` and kept the raised error message brief

daniel-makerx (approved):

achidlow (approved):

PR #327, opened 2024-10-17
Make CI/CD compile-all step check for untracked files 

achidlow:


.github/workflows/check-python.yaml
```diff
@@ -109,7 +109,7 @@ jobs:
           PYTHONUTF8: 1
 
       - name: Check diff
-        run: git diff --exit-code examples/ test_cases/
+        run: test -z "$(git status --porcelain examples/ test_cases/)"
 
   compile-all:
```
daniel-makerx, line 114: Should also update `check_for_diff` in `test_compile.py`

achidlow: Hol up let me just check this works first, first run should fail.

achidlow: Nah, same check fails on Windows due to line ending changes.
Since this is just to catch accidentally omitted files, I think we just have in the linux compile-all step.

.github/workflows/check-python.yaml
```diff
@@ -100,19 +102,28 @@ jobs:
           cache: "poetry"
 
       - name: Install dependencies
-        run: poetry install --no-interaction
+        run: poetry install --no-interaction --with cicd
```
daniel-makerx, line 105: I think we can remove --with cicd and Start LocalNet now

achidlow: Nice, done

daniel-makerx (approved):

PR #330, opened 2024-10-18
feat: add ARC-56 support

daniel-makerx:


src/puya/models.py
```diff
@@ -45,22 +45,32 @@ class ARC4CreateOption(enum.Enum):
...
-    return tuple((e1, e2) for (e1, e2) in elements)
-
-
 @attrs.frozen
-class ARC32StructDef:
+class ARC4StructField:
     name: str
-    elements: Sequence[tuple[str, str]] = attrs.field(default=(), converter=_freeze_list_of_lists)
+    type: str
+    struct: str | None
```
tristanmenzel, line 52: What's this field for? 🤔

daniel-makerx: ARC-56 is a backwards compatible extension of ARC-4, so `type` just defines the ARC4 encoding of an arg, which for a struct/named tuple will just be the tuple encoding with no field information. 
If the arg is actually a struct, then this field defines the name of the struct, which can be looked up on the Contract `structs` mapping to get the full information

See https://github.com/algorandfoundation/ARCs/pull/258/files#diff-6eb6fa5a0cc74e7fa2334541fb86699840a16e2518c98275de9268945c3dcaadR164 for more context

src/puya/arc56.py
```diff
@@ -0,0 +1,328 @@
...
+        self.aliases = dict[str, str]()
+        for struct in structs:
+            self.aliases[struct.fullname] = (
+                struct.fullname
+                if struct.name in self.aliases or struct.name in models.AVMType
+                else struct.name
+            )
+
+    def resolve[T: str | None](self, struct: T) -> T:
+        return self.aliases.get(struct, struct)  # type: ignore[arg-type, return-value]
```
achidlow, line 193:
```suggestion
    @typing.overload
    def resolve(self, struct: str) -> str: ...

    @typing.overload
    def resolve(self, struct: None) -> None: ...

    def resolve(self, struct: str | None) -> str | None:
        if struct is None:
            return None
        return self.aliases.get(struct, struct)
```

achidlow: Not pretty but it gets rid of the type ignore

achidlow (approved):

PR #331, opened 2024-10-18
Named tuple support

tristanmenzel:
## Proposed Changes

- Adds support for using native named tuples
- Named tuples are automatically encoded as ARC4 Structs when used on ABI boundaries
-


src/puyapy/awst_build/eb/compiled.py
```diff
@@ -43,82 +34,12 @@
 }
 
 
-class _LinearizedNamedTuple(NotIterableInstanceExpressionBuilder):
```
daniel-makerx, line 46: 🎉

daniel-makerx (approved):

PR #337, opened 2024-10-28
fix immutability for ARC4 struct and address types

daniel-makerx:
BREAKING CHANGE: modifying an `algopy.arc4.Address` will now raise an error


achidlow (dismissed):

achidlow (dismissed):

src/puya/awst/validation/immutable.py
```diff
@@ -0,0 +1,33 @@
+from puya import log
+from puya.awst import nodes as awst_nodes
+from puya.awst.awst_traverser import AWSTTraverser
+
+logger = log.get_logger(__name__)
+
+
+class ImmutableValidator(AWSTTraverser):
```
tristanmenzel, line 8: Should this include mutating nodes like `ArrayExtend` and `ArrayPop`?

daniel-makerx: Good point, initially this issue was addressing arc4.Address, which those nodes don't operate on. But we should still check regardless.

daniel-makerx:
Have found additional semantic compatibility issues around mutable arc4 types, which may require more breaking changes so converting to draft while addressing those

src/puyapy/awst_build/eb/arc4/_base.py
```diff
@@ -138,9 +138,8 @@ class _ARC4ArrayExpressionBuilder(BytesBackedInstanceExpressionBuilder[pytypes.A
     @typing.override
     def iterate(self) -> Expression:
         if not self.pytype.items.wtype.immutable:
-            logger.error(
-                "cannot directly iterate an ARC4 array of mutable objects,"
-                " construct a for-loop over the indexes via urange(<array>.length) instead",
+            logger.info(
+                "use `algopy.urange(<array>.length)` to iterate by index",
```
achidlow, line 142: why remove the fact that it's due to mutable values from the error message here?

achidlow: oh I see - maybe add a code comment that this condition creates an error during validation in puya core

daniel-makerx: Added comment

achidlow (approved):

PR #339, opened 2024-10-30
Refactor: Use const nodes for const arc4 values

tristanmenzel:
## Proposed Changes

- Use StringConstant(wtype=arc4_string_wtype, ...) for constant arc4 string values
- Use BoolConstant(wtype=arc4_bool_wtype, ...) for constant arc4 boolean values
- Update IR Builder to handle the trivial encoding of these constant values

## Why?
This is already how arc4 numeric types are handled. It makes the awst a little clearer to read and cuts down on the number of optimization passes required to output the same Teal.


src/puya/ir/builder/main.py
```diff
@@ -373,9 +373,24 @@ def visit_decimal_constant(self, expr: awst_nodes.DecimalConstant) -> TExpressio
...
+        match expr.wtype:
+            case wtypes.bool_wtype:
+                return UInt64Constant(
+                    value=int(expr.value),
+                    ir_type=IRType.bool,
+                    source_location=expr.source_location,
+                )
+            case wtypes.arc4_bool_wtype:
+                return BytesConstant(
+                    value=(128 if expr.value else 0).to_bytes(1),
```
daniel-makerx, line 385: It would be nice to keep ARC4 specific logic within the arc4 builder module, but not sure if it's worth it for this small bit of code. Maybe just have an ARC4_FALSE and ARC4_TRUE `bytes` constant in that module and use them here?

src/puya/ir/builder/main.py
```diff
@@ -394,6 +409,15 @@ def visit_string_constant(self, expr: awst_nodes.StringConstant) -> BytesConstan
             value = None
         if value is None or len(value) > algo_constants.MAX_BYTES_LENGTH:
             raise CodeError(f"invalid {expr.wtype} value", expr.source_location)
+        match expr.wtype:
+            case wtypes.string_wtype:
+                pass
+            case wtypes.arc4_string_alias:
+                value = len(value).to_bytes(2) + value
```
daniel-makerx, line 416: Again be nice to keep ARC4 stuff together, I think there is a function in arc4 module for encoding the length header, maybe use it here?

Also move the value length check after the match block to cover ARC4 values

test_cases/arc4_dynamic_arrays/out/DynamicArrayContract.approval.teal
```diff
@@ -463,7 +463,7 @@ get_string3:
     proto 0 1
     // arc4_dynamic_arrays/contract.py:185
     // return arc4.String("Hello World")
-    pushbytes 0x000b48656c6c6f20576f726c64
+    pushbytes "\x00\x0bHello World"
```
daniel-makerx, line 466: Since this value is a mix of UTF8 and escaped hex, might be better to keep the original base16 encoding.

The source comment provides the context of the UT8 value already

src/puya/ir/builder/main.py
tristanmenzel: I had a quick look at the arc4 builder stuff and any existing code there is aimed at writing the IR to do the encoding rather than encoding it at compile time. We could add something to do the encoding but 90% of it would look the same as this function which feels a little redundant to duplicate

daniel-makerx (approved):

PR #343, opened 2024-11-01
doc: update documentation regarding nested tuples, named tuples and immutability options

daniel-makerx:


docs/lg-types.md
```diff
@@ -230,8 +233,22 @@ In saying that, there are many places where built-in Python types can be used an
 
 ### tuple
 
-Python tuples are supported as arguments to subroutines, local variables, return types. Nested tuples
-are _not_ currently supported.
+Python tuples are supported as arguments to subroutines, local variables, return types.
+
+### typing.NamedTuple
+
```
achidlow, line 239: suggestions:
- link to `typing.NamedTuple` official docs here
- note that `typing.NamedTuple` can not be extended even in regular Python
- note that we don't support default values (yet?)

achidlow (approved):

PR #347, opened 2024-11-06
feat(internal): Add support for constant value default args in puya

tristanmenzel:
## Proposed Changes

- Small tweak to the default args schema to support passing constant value default args to the puya compiler
-
-

### DIFF
--- a/src/puya/ir/arc4_router.py
+++ b/src/puya/ir/arc4_router.py
@@ -11,6 +11,7 @@ from puya.awst import (
 )
 from puya.errors import CodeError, InternalError
 from puya.models import (
+    ABIMethodArgConstantDefault,
     ARC4ABIMethod,
     ARC4ABIMethodConfig,
     ARC4BareMethod,
@@ -474,7 +475,7 @@ def _validate_default_args(
         args_by_name = {a.name: a for a in method.args}
         for (
             parameter_name,
-            source_name,
+            default_source,
         ) in method.arc4_method_config.default_args.items():
             # any invalid parameter matches should have been caught earlier
             parameter = args_by_name[parameter_name]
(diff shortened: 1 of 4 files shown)


src/puya/ir/arc4_router.py
```diff
@@ -681,6 +693,18 @@ def _get_abi_signature(subroutine: awst_nodes.ContractMethod, config: ARC4ABIMet
     return f"{config.name}({','.join(arg_types)}){return_type}"
 
 
+def _is_valid_client_literal_for_arc4_type(literal: str | int, arc4_type_alias: str) -> bool:
```
daniel-makerx, line 696: [ARC-56](https://github.com/algorandfoundation/ARCs/pull/258/files#diff-6eb6fa5a0cc74e7fa2334541fb86699840a16e2518c98275de9268945c3dcaadR182) supports any ABI type, and I think ARC-32 did too, but the rules around encoding weren't clear.

So we should probably allow any constant value AVM / ABI value to be used in user code (which also means the default values should be part of AWST?). And then deal with how that value should be encoded into the spec within ARC-32 and ARC-56 code.

tristanmenzel: the ARC32 spec only lists int or str (not even bytes) but it's possible the clients actually support more than this.

daniel-makerx: Thats where its ambiguous in ARC-32, once encoded in to ARC-32/ ARC-56 it's only a str or int (i.e. it's encoded form in hex/base64 for ARC4 types), but the actual value in the program can be any high level type.

i.e. this should be valid (once python supports kwarg default values)

```python
class MyStruct(arc4.String, frozen=True):
    foo: arc4.String
    bar: arc4.String

@arc4.abimethod
def foo(
    self, 
    my_struct: MyStruct = MyStruct(arc4.String("foo"), arc4.String("bar")),
) -> None:
    ...
```

ARC-32
```ts
{
    "source": "constant", 
    /* data is UTF-8 encoded bytes for constant (format not defined in ARC-32)*/
    "data": "\\x00\\x04\\x00\\x09\\x03foo\\x03bar"
}
```

ARC-56 will be something like
```ts
{
      "source": "literal",
      /* data is base64 encoded bytes for constant*/
      "data": "AAQACQNmb28DYmFy",
      "type": "(string,string)"
}
```

achidlow:
Addressed by #371

PR #355, opened 2024-11-19
Add AWST Assert node

daniel-makerx:


src/puya/awst/nodes.py
```diff
@@ -261,6 +261,15 @@ def accept(self, visitor: StatementVisitor[T]) -> T:
         return visitor.visit_return_statement(self)
 
 
+@attrs.frozen
+class AssertStatement(Statement):
```
tristanmenzel, line 265: puya-ts requires an `AssertExpression` as typescript does not have an assert statement so assert is managed by a function call which can be used anywhere a void expression is valid. Presumably puyapy could just make use of ExpressionStatement(AssertExpression) so we would still only need one node

daniel-makerx: Made the node an AssertExpression instead

tristanmenzel (approved):
LGTM

PR #356, opened 2024-11-21
update mypy, replace `black` with `ruff format`

achidlow:
Other dev tools also updated


.pre-commit-config.yaml
```diff
@@ -1,15 +1,14 @@
 repos:
   - repo: local
     hooks:
-      - id: black
-        name: black
+      - id: ruff-format
+        name: ruff-format
         description: "Black: The uncompromising Python code formatter"
```
daniel-makerx, line 6: mmm... lies

daniel-makerx (approved):

PR #362, opened 2024-12-10
docs: add initial array adr

neilcampbell:
Collaboration with @daniel-makerx to create an Array ADR based on the discussions we've had thus far.

### DIFF
--- a/docs/architecture-decisions/2024-12-10_arrays.md
+++ b/docs/architecture-decisions/2024-12-10_arrays.md
@@ -29,7 +29,7 @@ The developer experience with Puya Arrays will depend on the API design (mutable
 
 -   [AlgoKit Guiding Principles](https://github.com/algorandfoundation/algokit-cli/blob/main/docs/algokit.md#guiding-principles) - specifically Seamless onramp, Meet developers where they are, Leverage existing ecosystem
 -   [Algorand Python Principles](https://algorandfoundation.github.io/puya/principles.html#principles) - specifically Least surprise and Leverage existing ecosystem by maintaining semantic compatibility.
--   [Algorand TypeScript Guiding Principles](https://github.com/algorandfoundation/puya-ts/blob/main/docs/README.md#guiding-principals)
+-   [Algorand TypeScript Guiding Principles](https://github.com/algorandfoundation/puya-ts/blob/main/docs/README.md#guiding-principals) - specifically TEALScript compatibility and minimise Algorand Python disparity.
 
 ## Options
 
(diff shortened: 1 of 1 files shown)


docs/architecture-decisions/2024-12-10_arrays.md
```diff
@@ -0,0 +1,237 @@
...
+-   An array of mutable elements is not limited to 4096 bytes, however each element would still have this limit
+
+**Cons**
+
+-   Easily possible to exhaust scratch slots quickly when working with an array of mutable elements as each element would require it's own slot
+-   It's the most complex option to fully implement
+
+## Preferred option
+
+In an ideal world being able to fully support option 4 would be great, however due to the constrained nature of the AVM it's not really feasible. The other options on their own will work, however there is some differences to what developers would be used to. As a result we feel that a combination of both option 2 and option 3 would give developers the familiarity and control they would expect.
```
joe-p, line 233:
```suggestion
In an ideal world being able to fully support option 4 would be great, however due to the constrained nature of the AVM it's not really feasible. The other options on their own will work, however there is some differences to what developers would be used to. As a result we feel that a combination of both option 2 and option 3 would give developers the familiarity and control they would expect. Having both options also allows front-ends to choose which implementation would make the most sense for native array behavior. 
```

joe-p (approved):
LGTM. Think this is a great encapsulation of the conversation! I have a small suggestion to capture the fact that this decision was made with front-end support in mind.

docs/architecture-decisions/2024-12-10_arrays.md
```diff
@@ -0,0 +1,237 @@
...
+
+-   Provide a more efficient array implementation for general purpose use
+-   Provide an API that communicates clearly the effect of operations and is semantically correct
+-   Support a broad range of types (including non ARC-4 types)
+
+## Principles
+
+-   [AlgoKit Guiding Principles](https://github.com/algorandfoundation/algokit-cli/blob/main/docs/algokit.md#guiding-principles) - specifically Seamless onramp, Meet developers where they are, Leverage existing ecosystem
+-   [Algorand Python Principles](https://algorandfoundation.github.io/puya/principles.html#principles) - specifically Least surprise and Leverage existing ecosystem by maintaining semantic compatibility.
+-   [Algorand TypeScript Guiding Principles](https://github.com/algorandfoundation/puya-ts/blob/main/docs/README.md#guiding-principals)
```
lempira, line 32: Thanks for putting this together @neilcampbell. This looks like it captured our discussions, options, and decision well. This is more of a comment re: the Algorand Typescript Guiding Principles. Similar to the Python guiding principles, there should probably a principle about maintaining semantic compatibility with Typescript.

lempira (approved):

docs/architecture-decisions/2024-12-10_arrays.md
neilcampbell: I'm happy to make this clearer. I guess the intent of the "Algorand Python" section within the guiding principles is to inherit the Algorand Python guiding principles and reduce duplication.

docs/architecture-decisions/2024-12-10_arrays.md
```diff
@@ -0,0 +1,237 @@
...
+// replace element
+arr = arr.with(0, 123);
+// remove elements
+arr = arr.slice(0, 1);
+// read element
+const value = arr[index];
+// pass to subroutine, does not require additional complexity to support this
+someReadonlyFunction(arr);
+// for a subroutine to return a modified array, it must be an explicitly returned value
+arr = someFunctionThatMutates(arr);
```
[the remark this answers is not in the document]
daniel-makerx: The value assigned to `arr` is immutable, but the variable `arr` is not defined as `const` so can be reassigned like this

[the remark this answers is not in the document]
tristanmenzel: `arr` is correct here. Being immutable does not prevent reassignment. For example strings and numbers are already immutable. It doesn't prevent you from writing `a = a + b` to update `a`.

PR #364, opened 2024-12-16
move lg-arc4 to docstrings

joe-p:
This is a draft PR showing what moving existing documentation to docstrings would look like. The main idea is that all of the conceptual documentation will be hosted on the new devportal, thus making a lot of the conceptual content redundant. The goal is to make docs easier to maintain and to provide more context in docstrings where applicable.


docs/lg-arc4.md
```diff
@@ -1,225 +0,0 @@
...
-        asset: Asset,
-        pay: gtxn.PaymentTransaction,
-        account: Account,
-        app: Application,
-        axfr: gtxn.AssetTransferTransaction        
-    ) -> None:
-        ...
-
-```
-### Mutability
```
robdmoore, line 216: Not all of the documentation on this page seems to have been replicated, e.g. this section (unless I'm missing something?)

robdmoore:
I like the idea of moving more docs into doc strings - it's powerful for automation and maintenance and also improves intellisense.

One thing I would caution though is removing conceptual documentation completely isn't a good experience since, per the UX report we did, it's important to support unknown information tasks as well as known information tasks. Completely removing the ARC-4 page and requiring people to navigate to the specific parts of it in the A-Z index won't work for that.

PR #369, opened 2025-01-09
fix: accept address or index param for VoterParamsGet methods

boblat:
## Proposed Changes

According to the [doc](https://github.com/algorand/go-algorand/blob/b2d41d0608b2a3e4c302df358950bdeb8b80be83/data/transactions/logic/TEAL_opcodes_v11.md#voter_params_get), the stack arg `A` of `voter_params_get` op code is an Account which is the same as the stack arg `A` of `acct_params_get` op code. 

This PR changes the type of parameter `a` for the methods of `VoterParamsGet` class to be `Account | UInt64 | int` to be inline with `AcctParamsGet` class.

### DIFF
--- a/langspec.puya.json
+++ b/langspec.puya.json
@@ -6889,7 +6889,7 @@
             "stack_inputs": [
                 {
                     "name": "A",
-                    "stack_type": "any",
+                    "stack_type": "address_or_index",
                     "doc": null
                 }
             ],
--- a/scripts/transform_lang_spec.py
+++ b/scripts/transform_lang_spec.py
@@ -260,6 +260,7 @@ def _patch_lang_spec(lang_spec: dict[str, typing.Any]) -> None:
         "asset_holding_get",
         "balance",
         "min_balance",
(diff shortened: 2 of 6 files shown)


achidlow:
The fix itself looks good but the stubs version will need to be bumped also. Take a look at the [Versioning section of the stubs README.md](https://github.com/algorandfoundation/puya/blob/main/stubs/README.md#versioning). 

The question is whether we treat this as a breaking change or as a bug-fix.

Technically, since there could exist contracts which will fail to type-check and thus compile after this change, it is a breaking change.
However, the stubs were actually incorrect before this fix, so from that lens, it's a bug-fix.
Additionally, this is a v11 op code, so not on main-net yet, so from that perspective there can't be any production that will break with this change.

boblat:
The stubs version has been bumped from `2.4.0` to `2.4.1` as the change is considered to be a bug fix and not a breaking change.

achidlow (approved):

PR #375, opened 2025-01-16
feat: update ops with AVM 11 changes

daniel-makerx:


achidlow (approved):

stubs/algopy-stubs/op.pyi
```diff
@@ -49,6 +49,20 @@ class ECDSA(str):
     secp256r1 curve, NIST standard
     """
 
+class MimcConfigurations(str):
```
achidlow, line 52: Should this be `MIMCConfiguratons`?

daniel-makerx: Yep

PR #400, opened 2025-02-26
docs: Front end development contribution guide & awst node documentation

tristanmenzel:
## Proposed Changes

-
-
-


docs/front-end-guide/02-designing-a-language.md
```diff
@@ -0,0 +1,40 @@
...
+
+The first step to adding a new front end for Puya is designing the language itself. 
+
+ - Will you build on top of an existing language (as Algorand Python and Algorand TypeScript do), or define an entirely new language from scratch
+   - Building on top of an existing language allows you to leverage existing developer tools and presents a familiar developer experience for your users; but limitations of the source langauge will constrain your solution.
+   - Defining a language from scratch gives you the ultimate in flexibility to build something fit for the purpose of writing smart contracts and logic signatures but represents a significantly larger undertaking. Developer experience will be heavily influenced by the effort put into accompanying developer tools (linter, formatter, language-service/autocomplete etc) 
+ - If building on an existing language, will you strive for semantic compatability with the source language - or match the language only in syntax. 
+   - Maintaining semantic compatability can be move restrictive than the alternative, but it ensures existing developer tooling such as code quick fixes are still relevant to code written in your dialect.
```
neilcampbell, line 13:
```suggestion
   - Maintaining semantic compatibility can be more restrictive than the alternative, but it ensures existing developer tooling such as code quick fixes are still relevant to code written in your dialect. 
```

docs/front-end-guide/02-designing-a-language.md
```diff
@@ -0,0 +1,40 @@
...
+
+Regardless of which option is chosen, the language will be composed of a set of supported syntax (likely a subset of the source language if based on an existing one), and a library of types and functions representing the full API of your solution. 
+
+### Primitive and compound types
+
+AWST defines a number of different primitive and compound types. You will need to decide how your language represents these types and how they interact. Notably your language will need to support a range of binary expressions for math operations on integers, and operations for slicing and dicing byte arrays. It should also define types for arrays and tuples.
+
+### Contract and logic signature paradigms
+
+Next you should decide how your language will represent contracts and logic signatures. Algorand Python and Algorand TypeScript use classes to represent contracts - _newing_ up a contract class is akin to creating the application with instance members use to define contract methods and state. For Algorand Python, logic signatures are implemented as a pure function whilst Algorand TypeScript uses a class again due to technical limitations of the language.
```
neilcampbell, line 23:
```suggestion
Next you should decide how your language will represent contracts and logic signatures. Algorand Python and Algorand TypeScript use classes to represent contracts - _newing_ up a contract class is akin to creating the application with instance members used to define contract methods and state. For Algorand Python, logic signatures are implemented as a pure function whilst Algorand TypeScript uses a class again due to technical limitations of the language. 
```

docs/front-end-guide/03-parsing.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Parsing
+
+### Generating AST
+
+If you have elected to extend an existing language then there is hopefully an existing compiler you can leverage for parsing. In the case of Algorand Python, it was built on top of [mypy](https://mypy-lang.org/) and makes use of the AST generated by invoking `mypy.build.build`^. Likewise, Algorand TypeScript is built on top of the TypeScript Compiler API and makes use of `ts.createProgram` to parse the source files into AST.
+
+> ^ Depending on when you are reading this, work is underway to replace the use of mypy in puyapy with cpython for performance reasons.
+
+If you are unfortunate enough to have chosen a language which does not an existing parser you can capitalise on, or you are going down the route of an entirely bespoke language then your option is to write your own lexers and parsers which is well beyond the scope of this guide. Luckily there are plenty of guides available on the internet.
```
neilcampbell, line 13:
```suggestion
If you are unfortunate enough to have chosen a language which does not have an existing parser you can capitalise on, or you are going down the route of an entirely bespoke language then your option is to write your own lexers and parsers which is well beyond the scope of this guide. Luckily there are plenty of guides available on the internet. 
```

docs/front-end-guide/readme.md
```diff
@@ -0,0 +1,17 @@
+# Building a front end for Puya
+
+This guide is for those looking to add support for a new front end language to Puya. It makes several assumptions about the reader.
+
+ - They are very familiar with building smart contracts and logic signatures on the Algorand Block Chain and are aware of the AVM and Teal op codes
+ - They are familiar with Puya's existing front end languages of Algorand Python and Algorand TypeScript
+ - They understand lexing and parsing, and how that applies to the language they would like to target
+
+Whilst making use of Puya's back end abstracts away the worst of the complexities in lowering a high level front end language to the stack based teal, adding a new front end is still a significant undertaking.
```
neilcampbell, line 9:
```suggestion
Whilst leveraging the Puya back end takes care of a lot of the complexities in lowering a high level front end language to the stack based teal language, adding a new front end is still a significant undertaking. 
```

src/puya/awst/nodes.py
```diff
@@ -199,6 +221,16 @@ def accept(self, visitor: StatementVisitor[T]) -> T:
 
 @attrs.frozen
 class IfElse(Statement):
+    """
+    A stand c-like if-then-else statement. Execution will continue to the following statement after
+    an if_branch or else_branch has been executed unless those branches contain a control op which
+    leads elsewhere (eg. return/break/continue)
```
achidlow, line 227:
```suggestion
    A standard C-like if-then-else statement. Control flow resumes at the next statement, unless interrupted e.g. with a `Goto` node.
```

src/puya/awst/nodes.py
```diff
@@ -199,6 +221,16 @@ def accept(self, visitor: StatementVisitor[T]) -> T:
...
+
+    condition: A boolean condition to be evaluated
```
achidlow, line 229: Field documentation should go on the fields - they can't technically be docstrings but most tooling (e.g. VSCode and PyCharm) will treat them as such.

src/puya/awst/nodes.py
```diff
@@ -209,6 +241,20 @@ def accept(self, visitor: StatementVisitor[T]) -> T:
 
 @attrs.frozen
 class Switch(Statement):
+    """
+    Evaluates `value` and the `key` expression of all cases.
```
achidlow, line 245: Whilst this is an important detail, the docstring should start with a high level summary of the functionality.

Suggestion:
> Compare an expression against a set of possible values, and if a match is found, execute the corresponding block. Otherwise, execute the default block (if supplied). Control flow resumes at the next statement, unless interrupted e.g. with a `Goto` node.

src/puya/awst/nodes.py
```diff
@@ -219,6 +265,16 @@ def accept(self, visitor: StatementVisitor[T]) -> T:
 
 @attrs.frozen
 class WhileLoop(Statement):
+    """
+    A pre-checked loop construct.
```
achidlow, line 269: What does this mean?

src/puya/awst/nodes.py
```diff
@@ -228,22 +284,33 @@ def accept(self, visitor: StatementVisitor[T]) -> T:
 
 @attrs.frozen
 class LoopExit(Statement):
-    """break out of the current innermost loop"""
+    """
+    Goto the statement immediately following the current innermost loop construct
```
achidlow, line 288: Maybe add something about generally being equivalent to a `break` statement in C-family languages.

src/puya/awst/nodes.py
```diff
@@ -228,22 +284,33 @@ def accept(self, visitor: StatementVisitor[T]) -> T:
...
+    """
 
     def accept(self, visitor: StatementVisitor[T]) -> T:
         return visitor.visit_loop_continue(self)
 
 
 @attrs.frozen
 class ReturnStatement(Statement):
+    """
+    Return from the current subroutine with an optional value.
```
achidlow, line 309: Maybe add that the value is only optional if the subroutine is typed as returning void.

src/puya/awst/nodes.py
```diff
@@ -252,6 +319,14 @@ def accept(self, visitor: StatementVisitor[T]) -> T:
 
 @attrs.frozen
 class AssertExpression(Expression):
+    """
+    Asserts the `condition` is true and errors with `error_message` if it is not. If no condition
```
achidlow, line 323: Probably needs to mention that the message is only a comment in the TEAL, so it can't be dynamic in any way, and relies on the caller to extract it from the source...

src/puya/awst/nodes.py
```diff
@@ -322,6 +418,11 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @enum.unique
 class BytesEncoding(enum.StrEnum):
+    """
+    Captures the source encoding of a bytes constant so that an appropriate encoding
+    can be used to produce a human-readable value in code output.
```
achidlow, line 423: it's more about preserving constant in it's original form/encoding, where possible.

src/puya/awst/nodes.py
```diff
@@ -360,6 +461,14 @@ def __repr__(self) -> str:
 
 @attrs.frozen(kw_only=True)
 class BytesConstant(Expression):
+    """
+    A compile-time bytes constant. Actual type could be any expression which is backed by the
+    bytes type.
```
achidlow, line 466:
```suggestion
    A compile-time bytes constant.
    
    The wtype is allowed to be anything which is ultimately backed by a single stack-value of byteslice, but front ends should prefer to use a more strongly typed node where possible, e.g. for a `bigint`, use an `IntegerConstant`.
```

src/puya/awst/nodes.py
```diff
@@ -396,6 +515,12 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @attrs.frozen
 class TemplateVar(Expression):
+    """
+    An expression whose value will be substituted with a constant value prior to deployment.
```
achidlow, line 519:
```suggestion
    An placeholder expression for a constant value which is not known at compile time, and must be substituted prior to deployment.
```

src/puya/awst/nodes.py
```diff
@@ -1123,6 +1271,10 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @attrs.frozen
 class BytesComparisonExpression(Expression):
+    """
+    Compare two bytes, or bytes backed values.
```
achidlow, line 1275: bytes-comparable is a subset of bytes-backed

src/puya/awst/nodes.py
```diff
@@ -1184,6 +1373,10 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @attrs.frozen
 class PuyaLibData:
+    """
+    Metadata for puya library functions.
```
achidlow, line 1377: This one should perhaps remain undocumented - not sure why it's not `_PuyaLibData` but I presume there was a reason.

src/puya/awst/nodes.py
```diff
@@ -1407,6 +1681,11 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @attrs.frozen
 class UInt64AugmentedAssignment(Statement):
+    """
+    Applies the specified binary operator to target and value and assigns the result to
+    target. Returns the result of the binary operation.
```
achidlow, line 1686: It's a statement, so there's no concept of "returns" here?

It's also an important detail, and the raison d'être of these nodes, that `target` is evaluated once and only once.

src/puya/awst/nodes.py
```diff
@@ -1454,6 +1743,10 @@ def accept(self, visitor: StatementVisitor[T]) -> T:
 
 @attrs.frozen
 class Emit(Expression):
+    """
+    Emits an ARC28 log event using the provided signature and struct
```
achidlow, line 1747: Hmmm, this node could probably use a bit of refinement - the signature should probably either be derived from value, or the value should be a sequence/mapping. But that's for a seperate issue. 

```suggestion
    Emits an ARC-28 log event using the provided signature, and the serialisation of the value.
```

src/puya/awst/nodes.py
```diff
@@ -1464,6 +1757,14 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @attrs.frozen
 class Range(Expression):
+    """
+    Returns an iterator-like expression which can be used in a ForInLoop. The iterator
+    emits uint64 values starting and start and incrementing by the step value until the result
+    is greater than or equal to the stop value.
+
+    The stop value is not emitted.
+    eg. range(0,4,2) emits [0, 2]
```
achidlow, line 1766:
```suggestion
    Produces a lazy-computed sequence which can be used with a `ForInLoop` (usage anywhere else is an error). 
    
    The sequence values come from the range `[start, stop)` (i.e. the range is inclusive of `start` and exclusive of `stop`). Step is the size of the steps within the range.
```

src/puya/awst/nodes.py
```diff
@@ -1475,6 +1776,14 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @attrs.frozen
 class Enumeration(Expression):
+    """
+    Returns an iterator-like expression which can be used in a ForInLoop. The iterator emits a
```
achidlow, line 1780: Similar vibe to `Range` here, it's a lazy-sequence, and only usable with `ForInLoop`.

src/puya/awst/nodes.py
```diff
@@ -1488,6 +1797,12 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @attrs.frozen
 class Reversed(Expression):
+    """
+    Returns an iterator-like expression which can be used in a ForInLoop. The iterator emits all
```
achidlow, line 1801: Similar vibe to Range here, it's a lazy-sequence, and only usable with ForInLoop.

src/puya/awst/nodes.py
```diff
@@ -1501,6 +1816,14 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @attrs.frozen
 class ForInLoop(Statement):
+    """
+    A loop construct which assigns the items Lvalue for each iterable item in sequence and
+    evaluates loop_body.
+
+    sequence: Any iterable expression
+    items: An lvalue expression which can be assigned before each iteration of the loop
```
achidlow, line 1824: can -> will ?

src/puya/awst/nodes.py
achidlow, line 1: Field documentation should go on the fields - they can't technically be docstrings but most tooling (e.g. VSCode and PyCharm) will treat them as such.

docs/front-end-guide/readme.md
achidlow, line 1: I think this should be `index.md`? Then you also don't need to specify it in the parent index.

src/puya/awst/nodes.py
tristanmenzel: The condition is checked before the loop body, as opposed to post checked ie `do {} while(condition)`

tristanmenzel: This is explained in the following paragraph though so maybe superfluous

src/puya/awst/nodes.py
tristanmenzel: Oh of course, in javascript it's an expression so I ended up not using these nodes

docs/front-end-guide/readme.md
tristanmenzel: 🤷

achidlow: Okay you still need to reference it as `/index` but it's the right name in this context. 
Have you checked what the build looks like? I'd suggest moving it down to the bottom of the index, since it's probably the most advanced part of the docs.
It also looks like there's issues with titling:
image

src/puya/awst/nodes.py
achidlow: I see you resolved this, but it looks like there's still some left? Examples: `IntegerConstant`, `AddressConstant`, all the members of the `SubroutineTarget` union.

docs/front-end-guide/readme.md
daniel-makerx: Fixed up titles and navigation

PR #402, opened 2025-02-27
fix: do not remove the phi node if it is used as a replacement for another register

boblat:


test_cases/regression_tests/jira_142.py
```diff
@@ -0,0 +1,20 @@
+from algopy import ARC4Contract, arc4, log, subroutine
+
+
+class MyStruct(arc4.Struct):
+    foo: arc4.UInt64
+
+
+class AContract(ARC4Contract):
```
achidlow, line 8: If you name this MyContract it'll produce a shorter entry in `sizes.txt`... lol
Which, looks like that hasn't been updated yet, so CI will fail in a few more minutes.

tests/test_execution.py
```diff
@@ -1472,3 +1472,27 @@ def test_loop_else(harness: _TestHarness) -> None:
 
 def test_regression_194(harness: _TestHarness) -> None:
     harness.deploy(TEST_CASES_DIR / "regression_tests" / "issue_194.py")
+
+
+def test_inlining(harness: _TestHarness) -> None:
+    def test() -> None:
```
achidlow, line 1478: This looks like a duplication of the test contract? You could deploy from the compilation result there, like in the test above. This won't actually test the method though - for that you'd probably want to add a test in `test_arc32.py` (which on reflection just now should probably be `test_excution_arc4.py` or similar, oh well).

test_cases/regression_tests/jira_142.py
boblat: `sizes.txt` has been updated. :)

tests/test_execution.py
boblat: The extra test has been removed.

achidlow (approved):

PR #404, opened 2025-02-27
feat: pyinstaller based binary support 

aorumbayev:
## Proposed Changes

- Added integration with binary build workflow in CD pipeline. Includes building on arm and amd ubuntu/mac, amd windows (compatible with arm), and testing manylinux based pyinstalled binaries against major linux distros using new portability tests introduced to pytest
- Connected binary artifact generation to GitHub releases
- Tested codesigning in the following tmp pr https://github.com/algorandfoundation/puya/actions/runs/13659830480/job/38233320310?pr=404


tests/test_puya_cli.py
```diff
@@ -0,0 +1,95 @@
...
+        cwd=VCS_ROOT,
+        check=False,
+        env=ENV_WITH_NO_COLOR,
+    )
+    if check:
+        assert result.returncode == 0, result.stdout
+    return result
+
+
+def test_run_no_args() -> None:
```
daniel-makerx, line 48: some tests were recently added to `test_cli.py` that already cover some of this functionality being tested here (and don't commit an awst.json to source). So I think we can migrate `test_run_no_args`, `test_run_version` and `test_run_with_invalid_options` to that file and then remove the rest and the `hello_world.awst.json`

.github/actions/build-binaries/linux/action.yaml
```diff
@@ -0,0 +1,165 @@
...
+    description: "The directory to write artifacts you want to publish"
+    required: true
+  with_codesign:
+    description: "Flag to determine if we should sign the binary"
+    required: true
+    type: boolean
+  gpg-signing-private-key:
+    description: "The GPG signing private key (base64 encoded)"
+    required: true
+  gpg-signing-private-key-password:
```
daniel-makerx, line 17: Should these be repo secrets rather than workflow inputs? (and similar question for other sensitive values in the actions for other platforms)

.github/workflows/cd.yaml
daniel-makerx, line 1: nitpick: can we remove the whitespace changes in this file?

.github/workflows/check-python.yaml
daniel-makerx, line 1: nitpick: whitespace changes

.github/workflows/cd.yaml
```diff
@@ -122,9 +124,17 @@ jobs:
 
       - name: Publish to GitHub
         if: steps.get_branch.outputs.branch == 'main' && !inputs.dry_run && inputs.publish_github
-        run: poetry run semantic-release $DRY_RUN publish
+        run: |
+          poetry run semantic-release $DRY_RUN publish
+          release_version="$(poetry run semantic-release print-version --current)"
+          echo "RELEASE_VERSION=$release_version" >> $GITHUB_ENV
```
daniel-makerx, line 130: could this just output the release_version straight to $GITHUB_OUTPUT? or does it need to be a 2 step process?

.github/actions/build-binaries/linux/action.yaml
neilcampbell: This is an action, so it can't directly access secrets

.github/workflows/build-binaries.yaml
```diff
@@ -0,0 +1,163 @@
...
+        # macos-13 is the last available Intel Mac runner (mac os 13)
+        # See https://github.com/actions/runner-images?tab=readme-ov-file#available-images
+        os: [ubuntu-22.04, ubuntu-22.04-arm, windows-latest, macos-13, macos-14]
+
+    steps:
+      - name: Set signing condition
+        id: signing
+        run: |
+          # Allow signing on:
+          # 1. Main branch non-PR events when production_release is true OR
```
neilcampbell, line 32: Is there any reason to disable signing when on `main` branch and `production_release` is false?

pyproject.toml
```diff
@@ -30,7 +30,7 @@ immutabledict = "^4.2.0"
 packaging = "^24.0"
 colorama = { version = "^0.4.6", markers = "sys_platform == 'win32'" }
 mypy_extensions = "^1.0.0" # vendored mypy dependency
-cattrs = "^24.1"
+cattrs = "^24.1"
```
neilcampbell, line 33: nitpick: remove space

.github/actions/build-binaries/linux/action.yaml
aorumbayev: @daniel-makerx yeah i cant access the secrets directly, hence they are passed as workflow inputs which gh takes care of masking when dealing with values from secrets.

.github/workflows/build-binaries.yaml
aorumbayev: This conditional migrated from algokit-cli, https://github.com/algorandfoundation/algokit-cli/pull/604 - in case of puya i suppose it makes sense to allow running codesign on main when prod release is false as well. Modified to allow running codesigning on any push to main

.github/workflows/cd.yaml
aorumbayev: @daniel-makerx In my case the changes are auto performed on save by prettier plugin. So had to include a header comment that disabled prettier in that file. Let me know if we want to globally disable prettier within vscode to perform any on-save actions on yamls in the .github folder.

.github/workflows/check-python.yaml
aorumbayev: @daniel-makerx same comment as in https://github.com/algorandfoundation/puya/pull/404#discussion_r1984027216, let me know if you are ok with adding prettier exclude rules to .vscode/setting.json -> that way i can restore without worrying about on-save behaviour

.github/workflows/cd.yaml
aorumbayev: Updated to output directly into github_output in 1 step. 
There were no particular reason for it other than it being a bit more explicit (having dedicated step).

tests/test_puya_cli.py
aorumbayev: @daniel-makerx slightly tweaked the code in test_cli to run puya from an env var that contains path to executable compiled via pyinstaller when running in ci cd. the test_puya_cli.py and artifact json file are removed now. Pytest mark is still used to run only the parametrized puya cli tests when testing binary executable in 'portability tests' step in ci

.github/actions/build-binaries/linux/action.yaml
```diff
@@ -0,0 +1,165 @@
...
+
+        # Skip the trust command as it requires user interaction
+        # Instead, just list the keys to confirm import
+        gpg --list-keys ${{ inputs.gpg-key-id }}
+
+        # Sign the package and checksum file
+        cd ${{ inputs.artifacts_dir }}
+
+        # Sign the main package
+        gpg --batch --pinentry-mode loopback --passphrase "${{ inputs.gpg-signing-private-key-password }}" \
```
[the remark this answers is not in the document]
aorumbayev: @neilcampbell good point, will patch the behaviour. Do we also want to codesign the tar as well? So it will have 3 sets of signatures for the binary itself, the _internal folder and the tar.gz containing both.

aorumbayev:
@daniel-makerx @neilcampbell most of the comments addressed. Just two minor follow up questions on the whitespace nitpicks + need to further test the codesigning behaviour on linux binaries as per Neil's comment (it was previously signing the tar.gz instead of the binary itself). Otherwise feel free to double check resolved comments if you agree with the way they got resolved

.github/actions/build-binaries/linux/action.yaml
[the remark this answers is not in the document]
aorumbayev: let me know what you think on latest.

1. It signs:
The main executable ("puya") with GPG
A checksum file of the "internal" folder contents with GPG
The SHA-256 checksum of the tar.gz archive with GPG

2. It uploads as artifacts:
The original executable package as a tar.gz
The SHA-256 checksum of the tar.gz file
The GPG signature of the SHA-256 checksum file
The GPG signature of the executable
The checksums file for the internal folder
The GPG signature of the internal checksums file

tests/test_cli.py
```diff
@@ -36,8 +36,13 @@ def run_puyapy(args: list[str | Path], *, check: bool = True) -> subprocess.Comp
     )
 
 
-def run_puya(args: list[str | Path]) -> subprocess.CompletedProcess[str]:
+def run_puya(
```
daniel-makerx, line 39: could this be simplified by setting the PATH before running the tests (so `shutil.which` picks up the binary) rather than introducing the extra environment variable?

.github/workflows/cd.yaml
daniel-makerx: I'm fine with having the prettier config options. 

The main motivation in general for not having the unnecessary white space changes is to preserve the git history as sometimes it can be useful to see when/why something was changed and formatting changes tend to get in the way of that.

tests/test_cli.py
aorumbayev: @daniel-makerx good point, further simplified - shutil will be picking up an external binary path when running portability tests in ci (instead of relying on explicit env var)

aorumbayev: @daniel-makerx refer to my dm in slack with details on why this isn't as trivial as it looks

aorumbayev: @daniel-makerx to confirm, instead it has explicit path being passed from 2 sets of tests (portability tests extract it from explicit env var, otherwise uses shutil.which). A few extra tweaks added to resolve issues with utf-8 output in windows binary.

scripts/distribution/package_windows.bat
```diff
@@ -1,2 +1,22 @@
 @echo off
-pyinstaller --clean --onedir --name puya --noconfirm src/puya/__main__.py --add-data ./src/puya/ir/_puya_lib.awst.json;puya/ir/ --exclude-module puyapy --exclude-module colorama --exclude-module mypy_extensions --optimize=2
+echo import os, sys, io > hook-puya.py
```
daniel-makerx, line 2: Would it be easier to check hook-puya.py into VCS instead of creating and deleting in a bat file?

nitpick: should also be `hook_puya.py`

tests/test_cli.py
```diff
@@ -36,17 +36,11 @@ def run_puyapy(args: list[str | Path], *, check: bool = True) -> subprocess.Comp
...
 
-def run_puya(
-    args: list[str | Path], *, use_binary: bool = True
-) -> subprocess.CompletedProcess[str]:
-    puya = shutil.which("puya")
-    if not puya and use_binary:
-        puya = os.environ.get("PUYA_BINARY")
-
-    assert puya is not None, "puya not found"
+def run_puya(args: list[str | Path], puya_path: str) -> subprocess.CompletedProcess[str]:
```
daniel-makerx, line 39: nitpick: puya_path should be first arg

.github/actions/build-binaries/linux/action.yaml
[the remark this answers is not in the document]
neilcampbell: Sounds good. I'm not sure there is much need for doing a checksum of the internal folder.
I'd imagine the main use case for the checksums is to ensure the downloaded tar.gz isn't altered in transit.

neilcampbell (approved):

.github/workflows/test-linux-distros.yaml
```diff
@@ -0,0 +1,250 @@
...
+
+          # Install poetry
+          RUN pip install poetry
+
+          # Create workspace directory
+          RUN mkdir -p /workspace
+          WORKDIR /workspace
+          EOF
+
+      - name: Build and push Docker image
```
achidlow, line 203: Why do we need to publish images to DockerHub to test linux binaries?

.github/workflows/build-binaries.yaml
```diff
@@ -0,0 +1,166 @@
...
+        shell: bash
+        run: |
+          artifacts_dir="${{ github.workspace }}${{ runner.os == 'Windows' && '\dist\artifacts' || '/dist/artifacts' }}"
+          mkdir -p $artifacts_dir
+          package_name_version="${{ inputs.release_version != '' && format('-{0}', inputs.release_version) || '' }}"
+          package_name="puya${package_name_version}-${{ runner.os }}_${{ runner.arch }}"
+          echo "PACKAGE_NAME=`echo $package_name | tr '[:upper:]' '[:lower:]'`" >> $GITHUB_ENV
+          echo "ARTIFACTS_DIR=${artifacts_dir}" >> $GITHUB_ENV
+
+      - name: Docker registry login
```
achidlow, line 55: Is this needed? A comment below says to run Python setup after a docker container has been run, but I can't see any between those two steps.

achidlow: Ah nevermind, it's in a separate action

pyproject.toml
```diff
@@ -40,6 +40,7 @@ mypy = "1.13.0"
 ruff = "^0.7.4"
 pre-commit = "^4.0.1"
 poethepoet = "^0.30.0"
+pyinstaller = { version = "^6.12.0", python = ">=3.12,<3.14" }
```
achidlow, line 43: Why a `dev` dependency instead of `cicd`?

.github/actions/build-binaries/linux/Dockerfile
```diff
@@ -0,0 +1,105 @@
+# Modified from https://github.com/oleksis/pyinstaller-manylinux
```
achidlow, line 1: Why? What's different?

.github/actions/build-binaries/linux/Dockerfile
```diff
@@ -0,0 +1,105 @@
...
+
+LABEL org.opencontainers.image.description="Run PyInstaller on ManyLinux 2.28 (AlmaLinux 8.7 based)"
+LABEL org.opencontainers.image.licenses=MIT
+
+SHELL ["/bin/bash", "-c"]
+
+ARG HOME=/root
+ARG PYTHON_VERSION=3.12
+ARG PYTHON_LAST=3.12.9
+ARG PYINSTALLER_VERSION=6.12.0
```
achidlow, line 14: This being hard coded here, and being version restricted but not hard coded in `pyproject.toml`, which is used for Windows and macOS, but not Linux, is a bit of a gotcha, no?

.github/workflows/test-linux-distros.yaml
aorumbayev: @achidlow its not pushing data to dockerhub though. Images are pushed to the repos ghcr and are only used for caching to speed up CI instead of building the images from scratch on each load.

.github/actions/build-binaries/linux/action.yaml
```diff
@@ -0,0 +1,218 @@
...
+        fi
+
+        # Set the cache tag based on architecture
+        ARCH=$(echo "${{ runner.arch }}" | tr '[:upper:]' '[:lower:]')
+        echo "CACHE_TAG=ghcr.io/${{ github.repository }}/puya-manylinux:${ARCH}" >> $GITHUB_OUTPUT
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Build and push Docker image
```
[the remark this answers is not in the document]
aorumbayev: @achidlow same reason as in https://github.com/algorandfoundation/puya/pull/404#discussion_r1986880057 - pushed to repos container registry to speed up workflow execution. I'd say even more relevant in case of testing on ~20x variations of linux distros that in the first case when its doing the build on arm/amd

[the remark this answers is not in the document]
aorumbayev: @achidlow i assume you also referred specifically to combination of flags `push` being set to true and `cache-to` being also specified. I'll see if it's possible to remove the push flag while still using `cache-to` - which should cache the intermediate layers only without pushing the finale image.

pyproject.toml
achidlow: We keep the dev dependencies light weight because they're installed by default.

PyInstaller as part of the build & release workflow belongs in the CI/CD group for sure, if it belongs in this file at all it might be best to use latest by default with option to pin, and in which case the pin should propagate to all builders, not just Linux or macOS+Windows.

.github/workflows/test-linux-distros.yaml
aorumbayev: I think i got what you mean though, I'll see if it's possible to remove the push flag while still using cache-to - which should cache the intermediate layers only without pushing the finale image.

.github/actions/build-binaries/linux/Dockerfile
```diff
@@ -0,0 +1,105 @@
...
+ARG HOME=/root
+ARG OPENSSL_DIR=/usr/local/ssl
+ARG PYENV_ROOT="${HOME}/.pyenv"
+
+ENV PATH="${PYENV_ROOT}/bin:${PYENV_ROOT}/shims:${OPENSSL_DIR}/bin:$PATH" \
+    PYENV_ROOT="${PYENV_ROOT}" \
+    HOME=${HOME}
+
+# Install minimal runtime dependencies
+RUN dnf -y install zlib bzip2 readline sqlite tk libffi libnsl2 libuuid gdbm && \
```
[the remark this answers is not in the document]
aorumbayev: @neilcampbell When Python is compiled from source (as it is here via pyenv), it requires SQLite development headers to build this module

.github/workflows/test-linux-distros.yaml
achidlow: Yeah nice, I saw how much it speeds things up when I tried to remove it lol

.github/actions/build-binaries/linux/Dockerfile
achidlow: Looks like you've updated it a bunch and simplified, which is good - I wonder at that point if it's worth even having that link in there? Comparing the two files they're quite different at this point.

.github/actions/build-binaries/linux/Dockerfile
achidlow: I think I found a way around this in my branch, we can defer installing PyInstaller until we actually run the container, then we can install it using poetry

.github/workflows/test-linux-distros.yaml
achidlow: I've managed to remove the need for docker at all when testing in my branch, will PR into this one once I'm done

.github/actions/build-binaries/linux/action.yaml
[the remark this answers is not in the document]
achidlow: Yeah nice, I also saw how much it speeds things up when I tried to remove it lol

.github/actions/build-binaries/linux/Dockerfile
aorumbayev: @achidlow sounds good, once you'll finalize your changes on the pr it might be easier for you to retarget against this branch then merge it here (would also be easier to see the diffs on what your pr simplifies). And then we can merge this pr (or apply any remaining last minute feedback/tweaks)

pyproject.toml
aorumbayev: @achidlow sounds good, i'll move it to cicd group after changes from your branch are applied

.github/workflows/cd.yaml
```diff
@@ -109,36 +112,123 @@ jobs:
...
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          packages-dir: stubs/dist
+          packages-dir: release-artifacts/stubs/dist
 
   publish-docs:
     name: Publish Docs
-    if: ${{ !inputs.dry_run && inputs.publish_docs }}
+    needs: [prepare-release]
+    if: needs.prepare-release.outputs.branch_name == 'main' && !inputs.dry_run && inputs.publish_docs
```
aorumbayev, line 231: @daniel-makerx @achidlow do we want to keep it as is or given new sequential flow -> run it only after all of the previous steps are succeeded ? Kept it as is given that in case of a failed release documentation can still be reverted via separate workflow trigger

aorumbayev:
@achidlow @daniel-makerx you can see full e2e tests in a forked repo https://github.com/aorumbayev/puya/actions/runs/13794601982 https://github.com/aorumbayev/puya/releases/tag/v4.6.6. Note that i found an alternative to using ghcr with buildx action - it had native support for github actions cache -> this eliminated the need to push entire docker image to ghcr and related permissions issues. Unfortunately i don't have access control rights to remove all of the following cached images from https://github.com/orgs/algorandfoundation/packages?repo_name=puya -> once pr is merged we can clean them up, if neither can access the delete options on package i'll reach out to David to perform a cleanup.

@neilcampbell as discussed in slack i removed checksum generation for _internal folder, gpg signing is performed on sha256, tar and the binary itself

.github/workflows/cd.yaml
```diff
@@ -109,36 +112,123 @@ jobs:
...
+  publish-to-github:
+    name: Publish to GitHub
+    needs: [prepare-release, build-and-upload-binaries]
+    if: needs.prepare-release.outputs.branch_name == 'main' && !inputs.dry_run && inputs.publish_github
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: write
+    env:
+      DRY_RUN: ${{ inputs.dry_run && '--noop' || '' }}
```
daniel-makerx, line 152: Since this job is now guarded entirely by `inputs.dry_run` can remove this env var

.github/workflows/cd.yaml
```diff
@@ -109,36 +112,123 @@ jobs:
...
+        with:
+          python-version: "3.12"
+          cache: "poetry"
+
+      - name: Install dependencies
+        run: poetry install --no-interaction --with cicd
+
       - name: Publish to GitHub
-        if: steps.get_branch.outputs.branch == 'main' && !inputs.dry_run && inputs.publish_github
         run: poetry run semantic-release $DRY_RUN publish
```
daniel-makerx, line 172: as per above can remove $DRY_RUN

scripts/distribution/hook_puya.py
```diff
@@ -0,0 +1,24 @@
+import io
+import os
+import sys
+
+# Set environment variables
+os.environ["PYTHONIOENCODING"] = "utf-8"
```
daniel-makerx, line 6: What does PyInstaller do with this script, and do we need these 3 environment variables and the stdout/stdin patching below? Typically for windows we just need `PYTHONUTF8` set before the process is launched to ensure it enters [UTF-8 mode](https://docs.python.org/3/library/os.html#python-utf-8-mode) but I figure this gets more complicated with whatever PyInstaller is doing?

.github/actions/build-binaries/linux/action.yaml
```diff
@@ -0,0 +1,235 @@
...
+      uses: docker/build-push-action@v6
+      env:
+        DOCKER_BUILD_SUMMARY: false
+      with:
+        context: ${{ github.action_path }}
+        file: ${{ github.action_path }}/Dockerfile
+        build-args: |
+          BASE_IMAGE=${{ steps.set-base-image.outputs.BASE_IMAGE }}
+          PYTHON_VERSION=${{ inputs.python_version }}
+        push: false
```
daniel-makerx, line 64: Since it isn't pushing now, rename this step?

.github/workflows/check-python.yaml
daniel-makerx, line 1: nitpick: Looks like prettier has struck again?

scripts/distribution/hook_puya.py
aorumbayev: @daniel-makerx this was a fix for the portability tests that were initially run via pytest before Adam's further refinements, 

The error i was getting was related to unicode chars:
```
def _run(args: list[str], *, check: bool = True) -> subprocess.CompletedProcess[str]:
        result = subprocess.run(
            args,
            text=True,
            encoding="utf-8",
            # capture stdout
            stdout=subprocess.PIPE,
            # redirect stderr to stdout, so they're interleaved in the correct ordering
            stderr=subprocess.STDOUT,
            cwd=VCS_ROOT,
            check=False,
            env=ENV_WITH_NO_COLOR,
        )
        if check and result.returncode != 0:
>           pytest.fail(f"{args[0]} failed: {result.stdout}")
E           Failed: D:\a\puya\puya\dist\puya\puya.exe failed: debug: Building IR for function _puya_lib.util.ensure_budget
E           debug: Sealing block@0
E           debug: Terminated block@0
E           debug: Looking for 'required_budget_with_buffer' in an unsealed block creating an incomplete Phi: block@1
E           debug: Created Phi assignment: let required_budget_with_buffer#1: uint64 = undefined while trying to resolve 'required_budget_with_buffer' in block@1
E           debug: Terminated block@1
E           debug: Sealing block@2
E           debug: Looking for 'fee_source' in an unsealed block creating an incomplete Phi: block@1
E           debug: Created Phi assignment: let fee_source#1: uint64 = undefined while trying to resolve 'fee_source' in block@1
E           debug: Terminated block@2
E           debug: Sealing block@3
E           debug: Terminated block@3
E           debug: Sealing block@4
E           debug: Terminated block@4
E           debug: Sealing block@5
E           debug: Terminated block@5
E           debug: Sealing block@6
E           debug: Terminated block@6
E           debug: Sealing block@1
E           debug: Traceback (most recent call last):
E             File "puya\errors.py", line 36, in log_exceptions
E             File "puya\ir\context.py", line 114, in log_exceptions
E             File "puya\ir\builder\main.py", line 127, in build_body
E             File "puya\awst\nodes.py", line 183, in accept
E             File "puya\ir\builder\main.py", line 1059, in visit_block
E             File "puya\awst\nodes.py", line 226, in accept
E             File "puya\ir\builder\main.py", line 1071, in visit_while_loop
E             File "puya\ir\builder\flow_control.py", line 150, in handle_while_loop
E             File "contextlib.py", line 144, in __exit__
E             File "puya\ir\builder\blocks.py", line 124, in activate_open_block
E             File "puya\ir\builder\blocks.py", line 241, in _seal_block_if_unlabelled
E             File "puya\ir\ssa.py", line 174, in seal_block
E             File "puya\ir\ssa.py", line 130, in _add_phi_operands
E             File "puya\log.py", line 329, in debug
E             File "puya\log.py", line 406, in _report
E             File "structlog\_native.py", line 169, in log
E             File "structlog\_base.py", line 215, in _proxy_to_logger
E             File "structlog\_output.py", line 110, in msg
E             File "encodings\cp1252.py", line 19, in encode
E           UnicodeEncodeError: 'charmap' codec can't encode character '\u03c6' in position 100: character maps to <undefined>
E           critical: UnicodeEncodeError: 'charmap' codec can't encode character '\u03c6' in position 100: character maps to <undefined>
```
And was due to the env vars set in pytest not being applied to python bundled inside the binary so i had to add a pyinstaller [runtime hook](https://pyinstaller.org/en/stable/when-things-go-wrong.html#changing-runtime-behavior) - which get code injected into the bundled application during the build process, but are designed to execute when the application runs. The issue was resolved by forcing the utf-8 on stdin and stdout while remaining env vars were added as extra safe guards. I can't easily recreate the original error at this point unless i can go back to the previous approach with pytest. So i'll removed the env vars, expect keeping the PYTHONUTF8 env var in place along with stdin/stdout patching in-place (as this was what fixed the initial issue in pytest)

.github/actions/build-binaries/windows/action.yaml
```diff
@@ -0,0 +1,109 @@
...
+          --noconfirm src/puya/__main__.py \
+          --add-data './src/puya/ir/_puya_lib.awst.json:puya/ir/' \
+          --exclude-module mypy_extensions \
+          --optimize=2 \
+          --hidden-import colorama \
+          --runtime-hook scripts/distribution/hook_puya.py
+
+    - name: Sign executable
+      if: ${{ inputs.with_codesign == 'true' }}
+      uses: azure/trusted-signing-action@v0.3.20
```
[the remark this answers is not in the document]
aorumbayev: bumped and tested against latest, all good

.github/actions/build-binaries/windows/action.yaml
```diff
@@ -0,0 +1,109 @@
...
+    required: true
+    type: string
+
+runs:
+  using: "composite"
+  steps:
+    - name: Configure build environment
+      shell: bash
+      run: |
+        echo 'C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x64' >> $GITHUB_PATH
```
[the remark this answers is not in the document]
aorumbayev: removed and tested that it works without it

scripts/distribution/hook_puya.py
achidlow: I've removed the hook, trimmed it down to the required code, and it'll now be done consistently regardless of binary on not. Also, disabled forcing NO_COLOR with the binary option.

aorumbayev:
@achidlow cache issue is resolved, after further reading on [`scope`](https://docs.docker.com/build/cache/backends/gha/#scope) - turned out that cache tag was not sufficient to ensure unique cache per architecture - had to add an explicit scope parameter in cache-to and cache-from fields:
```
cache-from: type=gha,scope=${{ runner.arch }}-build-cache
cache-to: type=gha,mode=max,scope=${{ runner.arch }}-build-cache
```

Retriggered pipeline a few times and did an extra commit to confirm its now reusing unique caches per architecture

achidlow (approved):

PR #409, opened 2025-03-04
fix: ensure bool is returned for all calls to `box_del` op code

boblat:


src/puya/awst/nodes.py
```diff
@@ -1583,7 +1583,7 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 @attrs.frozen
 class StateDelete(Expression):
     field: StorageExpression
-    wtype: WType = attrs.field(default=wtypes.void_wtype, init=False)
+    wtype: WType = attrs.field(default=wtypes.void_wtype)
```
daniel-makerx, line 1586: given the change in the IR building, it seems this should always default a `bool_wtype`?

src/puya/ir/builder/storage.py
```diff
@@ -89,17 +89,28 @@ def visit_state_get_ex(
 def visit_state_delete(
     context: IRFunctionBuildContext, statement: awst_nodes.StateDelete
 ) -> ValueProvider | None:
-    match statement.field:
-        case awst_nodes.BoxValueExpression(key=awst_key):
+    match statement:
+        case awst_nodes.StateDelete(
+            field=awst_nodes.BoxValueExpression(key=awst_key), wtype=wtypes.bool_wtype
```
daniel-makerx, line 94: as per AWST validation suggestion, this should also allow void

src/puyapy/awst_build/eb/storage/_common.py
```diff
@@ -78,7 +79,9 @@ class BoxValueExpressionBuilder(ValueProxyExpressionBuilder[pytypes.PyType, BoxV
     @typing.override
     def delete(self, location: SourceLocation) -> Statement:
         return ExpressionStatement(
-            expr=StateDelete(field=self.resolve(), source_location=location)
+            expr=StateDelete(
+                field=self.resolve(), source_location=location, wtype=wtypes.bool_wtype
```
daniel-makerx, line 83: can revert this change if we allow void with boxes

src/puya/awst/nodes.py
```diff
@@ -1583,7 +1583,15 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
...
+
+    @wtype.validator
+    def _validate_wtype(self, _: object, value: WType) -> None:
+        expected = (
+            wtypes.bool_wtype if isinstance(self.field, BoxValueExpression) else wtypes.void_wtype
```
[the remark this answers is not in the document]
boblat: Thanks Dan. It has been updated to allow void as well for BoxValueExpression.

daniel-makerx (approved):

PR #411, opened 2025-03-06
Fix/array encoding

tristanmenzel:
## Proposed Changes

- Allow for arrays of biguint
- Allow for arrays of arc4.Bool
- Allow arc4 static array to decode into tuple or native array


src/puya/awst/wtypes.py
```diff
@@ -389,6 +382,9 @@ def _name(self) -> str:
 
     @arc4_name.default
     def _arc4_name(self) -> str:
+        # Homogenous arc4 tuples of T are the same as T[N]
+        if all(t == self.types[0] for t in self.types[1:]):
+            return f"{self.types[0].arc4_name}[{len(self.types)}]"
```
tristanmenzel, line 387: This isn't needed but it does save `uint[20]` from turning into `[uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint, uint]`

daniel-makerx: I'm not sure if this is safe to do.

There are a few spots where two types are compared to see if they have equivalent ARC4 encodings by looking to see if their ARC-4 names are equal. I think this might interfere with that. Additionally there are some semantic differences between a homologous tuple of size N and a static array of size N. Mutability being the main one

src/puya/ir/types_.py
```diff
@@ -379,6 +379,8 @@ def wtype_to_encoded_ir_type(
                 for e in wtype.types
             ]
         )
+    if wtype == wtypes.biguint_wtype:
```
daniel-makerx, line 382: I don't think we can do this here because biguint values have variable sizes

src/puya/awst/wtypes.py
```diff
@@ -190,14 +190,7 @@ class NativeArray(WType, abc.ABC):
...
             raise CodeError("array element type cannot be void", self.source_location)
-        if element_type == arc4_bool_wtype:
-            # technically we could support this, but it requires a additional complexity to support
-            # and is less efficient than using a native bool
-            logger.error(
-                "arrays of arc4 bools are not supported, use an array of native bools instead",
-                location=self.source_location,
-            )
-        elif not element_type.immutable:
+        if not element_type.immutable:
```
daniel-makerx, line 193: Originally this guard was because a native array of arc4 bools is not bit packed and it would need to be for this to work correctly.

If it is indeed working correctly now then would be good to have some tests confirming this

src/puya/awst/wtypes.py
```diff
@@ -471,6 +467,15 @@ def _name(self) -> str:
     def _arc4_name(self) -> str:
         return f"{self.element_type.arc4_name}[{self.array_size}]"
 
+    @typing.override
+    def can_encode_type(self, wtype: WType) -> bool:
```
daniel-makerx, line 471: A test for this would be good, `abi_call` should work as a way to have a python test contract that converts from a native tuple to an ARC4 static array. 

I imagine TS can actually describe this sort of conversion formally in the type system?

src/puya/ir/types_.py
tristanmenzel: If I'm interpreting this method correctly it's asking what the ir type of an arc4 encoded version of the input type is - which is a constant size

src/puya/awst/wtypes.py
tristanmenzel: Why does the in memory representation of `arc4.Bool[]` matter? As long as we're not blindly converting StackArray values as arc4 arrays

src/puya/ir/types_.py
daniel-makerx: Unfortunately not, it represents the encoded IR type used in ReferenceArrays, which can only hold fixed size elements. A biguint would first need to be padded for this to work

src/puya/awst/wtypes.py
daniel-makerx: It works for `StackArray` but not `ReferenceArray` (as it doesn't do bitpacking, by doing this it means that the length of a reference array can easily be derived from `len(array.bytes) / element_size`  ) we added the guard for both types however to have a level of consistency between the two, and based on the assumption that there wouldn't be a great need for native arrays of `arc4.Bool`

The reason the encoding matters is we allow conversion from ReferenceArray to StackArrays and currently this is as simple as prepending the u16 length to the ReferenceArray (to make it ARC-4)

src/puya/ir/types_.py
daniel-makerx: I can see StackArrays aren't working with biguint, but they should so working on a fix for that

daniel-makerx:
The fixes in this PR and some additional ones discovered during testing are now in #424

PR #415, opened 2025-03-11
feat: puya daemon mode 

aorumbayev:
## Proposed Changes

- Draft implementation of daemon server for use in language server based on websockets and https://www.jsonrpcserver.com/en/stable/
- Extending cli interface for explicit compile/daemon commands while logging related args are available as global args at the root

## TODO

- Add tests for daemon mode in binary mode (requires binary support pr merged first)


.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
...
+            "type": "python",
+            "request": "launch",
+            "program": "${file}",
+            "purpose": [
+                "debug-test"
+            ],
+            "console": "integratedTerminal",
+            "justMyCode": false,
+            "env": {
+                "PYTEST_ADDOPTS": "-xvs -n 1"
```
aorumbayev, line 15: mainly to make debugging via test explorer in vscode more convenient - otherwise it will launch tests with default xdist -n auto. `justMyCode` also allows stepping into internal methods - often useful for debugging

aorumbayev:
moving discussion to miro

PR #423, opened 2025-03-18
feat: puya service mode [DRAFT]

aorumbayev:
## Proposed Changes

- Draft includes Dan's puya-service branch changes + latest main with pyinstaller support
- Adds cancellation mechanism into puyad server
- Sets 2 threads by default while allowing arg to pass custom amounts of workers
- Initial integration tests to test sucessful analyse endpoint call vs cancelled process
-


src/puya/puyad.py
```diff
@@ -0,0 +1,209 @@
...
+import logging
+import os
+import signal
+import sys
+import time
+import types
+from collections.abc import Callable
+from importlib.metadata import version
+from pathlib import Path
+from typing import Any, BinaryIO, Literal, NoReturn
```
daniel-makerx, line 12: nitpick: for the `typing` module specifically we have the convention in puya to not import individual members but always access via `typing.`

src/puya/__main__.py
```diff
@@ -38,12 +41,31 @@ def cli() -> None:
         choices=list(LogFormat),
         default=LogFormat.default,
     )
-    parser.add_argument("--options", type=Path, required=True)
-    parser.add_argument("--awst", type=Path, required=True)
+    parser.add_argument("--options", type=Path, required=False)
+    parser.add_argument("--awst", type=Path, required=False)
     parser.add_argument("--source-annotations", type=Path)
+    parser.add_argument("--service", action="store_true", help="Run in service mode")
+    parser.add_argument(
```
daniel-makerx, line 48: Do we need the service threads for initial version?

src/puya/puyad.py
```diff
@@ -0,0 +1,209 @@
...
+    def __init__(
+        self,
+        protocol_cls: type[JsonRPCProtocol] = PuyaProtocol,
+        converter_func: Callable[..., Any] = get_converter,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(protocol_cls, converter_func, **kwargs)
+        self._exit_handler_installed = False
+
+    def install_signal_handlers(self) -> None:
```
daniel-makerx, line 99: How does this different from the default behaviour with JsonRPCServer?

src/puya/puyad.py
```diff
@@ -0,0 +1,209 @@
...
+                    location=None,
+                )
+            ],
+        )
+
+        try:
+            with (
+                logging_context() as log_ctx,
+                contextlib.suppress(PuyaExitError),
+                log_exceptions(),
```
daniel-makerx, line 158: was the try/except required here? the `log_exceptions` should be catching any errors and including them on the log_ctx

src/puya/puyad.py
```diff
@@ -0,0 +1,209 @@
...
+            return AnalyseResult(
+                logs=[Log(level=LogLevel.error, message=f"Error: {e}", location=None)],
+            )
+
+        return result
+
+    return server
+
+
+def main() -> None:
```
daniel-makerx, line 190: can probably remove this if `puya` is going to be the main entry point?

tests/test_puyad.py
```diff
@@ -0,0 +1,298 @@
...
+from pathlib import Path
+from typing import Any
+
+from puya.log import Log, LoggingContext, LogLevel, logging_context
+from puyapy.compile import compile_to_teal
+from puyapy.options import PuyaPyOptions
+from tests import EXAMPLES_DIR
+
+
+class PuyadTestClient:
```
daniel-makerx, line 16: Can we use the JSONRPCClient similar to here https://github.com/algorandfoundation/puya/pull/423/files#diff-d45dabeb66ffac4280be491c788d11f23e4ea99f57ef3a2bd5718075b2f926ddR41

Also once you have that approach working in the tests, revert the puyapy/lsp/server.py change since they were just a quick example to make sure it worked initially

daniel-makerx: Should also be able to remove the need for a temporary file by following the example mentioned above

tests/test_puyad.py
```diff
@@ -0,0 +1,298 @@
...
+                    self.debug_print(f"Error stopping daemon: {e}")
+
+            self.is_running = False
+
+        for thread in [self.stdout_thread, self.stderr_thread]:
+            if thread and thread.is_alive():
+                thread.join(timeout=1)
+
+
+def generate_awst_file(
```
daniel-makerx, line 232: Can use `get_awst_cache(EXAMPLES_DIR)` to get the AWST for the examples dir and then filter to get the `examples.hello_world_arc4.contract.HelloWorldContract` AWST, this should be faster (as it caches the slow mypy portion) and also avoid the need for writing to disk

src/puya/__main__.py
```diff
@@ -38,12 +41,31 @@ def cli() -> None:
         choices=list(LogFormat),
         default=LogFormat.default,
     )
-    parser.add_argument("--options", type=Path, required=True)
-    parser.add_argument("--awst", type=Path, required=True)
+    parser.add_argument("--options", type=Path, required=False)
+    parser.add_argument("--awst", type=Path, required=False)
     parser.add_argument("--source-annotations", type=Path)
+    parser.add_argument("--service", action="store_true", help="Run in service mode")
```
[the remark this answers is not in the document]
aorumbayev: keeping global args, but moved the service under a command named `serve` which sounds like a slightly better semantics for the command. lmk what you think though, can revert back to `service` if needed

src/puya/puyad.py
aorumbayev: good point, addressed

src/puya/__main__.py
aorumbayev: removed - 2 should be sufficient and we can always increase or re introduce the flag if needed

src/puya/puyad.py
aorumbayev: in case of lsp it relies on LanguageServer (which is a subclass of BaseLanguageServer which is a subclass of JsonRPCServer) - puya service subclasses JsonRPCServer directly. Lsp relies on event loop, puya service uses threaded decorators from pygls, due to that sigint and sigterm signals are not working properly out of the box - this isnt an issue in the lsp server case (you can see that a single ctrl+c terminates it properly, probably due to the fact it doesn't leverage multi threading).

src/puya/puyad.py
aorumbayev: removed - do we keep it in pyproject executable entrypoints or remove it from there as well?

tests/test_puyad.py
aorumbayev: good point, refactored to leverage `JSONRPCClient`

aorumbayev: also had to explicitly add pytest asyncio dependency to leverage async without dealing with event loop setup manually

tests/test_puyad.py
aorumbayev: addressed

src/puyapy/lsp/server.py
```diff
@@ -34,8 +38,22 @@ def __init__(self, name: str, version: str) -> None:
         self.diagnostics = dict[str, tuple[int | None, list[types.Diagnostic]]]()
         self.analysis_prefix: Path | None = None
         self.current_refresh_token = object()
-
-    def parse_all(self) -> None:
+        self._puyad = JsonRPCClient(PuyaProtocol, get_converter)
```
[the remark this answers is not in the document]
daniel-makerx: Yes, we should drop the commit that added it. I just added it as an temporary example to verify it actually worked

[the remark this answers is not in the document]
daniel-makerx: https://github.com/algorandfoundation/puya/pull/423#discussion_r2015839766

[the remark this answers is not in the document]
neilcampbell: Ok cool 👍

neilcampbell (approved):

PR #424, opened 2025-03-18
Various ARC-4 encoding fixes

daniel-makerx:


src/puya/awst/wtypes.py
```diff
@@ -286,16 +279,32 @@ def name_to_index(self, name: str, source_location: SourceLocation) -> int:
...
-        return wtype == self.native_type
+        return self.native_type == wtype
+
+    @typing.override
+    def can_encode_type_in_array(self, wtype: WType) -> bool:
+        return wtype in (self, self.native_type)
 
 
-arc4_bool_wtype: typing.Final = ARC4Type(
+arc4_bool_wtype: typing.Final = _SimpleARC4Type(
```
tristanmenzel, line 307: This is a breaking change to the AWST for not a lot of value. Can we just have a default implementation on `ARC4Type` which is overridden by more complex types? 

If not, `_SimpleARC4Type` should be renamed to `SimpleARC4Type` as it's not really an internal implementation detail.

daniel-makerx: Ah good point, wanted to move away from a default implementation as it isn't always correct and so it's better to force new types to provide their own implementation. But didn't consider the implications of this change on serialization.

Have gone back to a concrete ARC4Type for now

src/puya/awst/wtypes.py
```diff
@@ -471,6 +509,29 @@ def _name(self) -> str:
...
+                return self.element_type.can_encode_type(single_type)
+        return isinstance(wtype, StackArray) and self.element_type.can_encode_type_in_array(
+            wtype.element_type
+        )
+
+    @typing.override
+    def can_encode_type_in_array(self, wtype: WType) -> bool:
+        if wtype in (self.native_type, self):
+            return True
+        return isinstance(wtype, ARC4StaticArray) and self.element_type.can_encode_type_in_array(
```
achidlow, line 531: Is this meant to check for `StackArray`?

achidlow: Yes, but code deleted anyway.

achidlow (changes requested):
Looking good I think, but let's add some more tests before merging - particularly negative tests.

achidlow (approved):

PR #427, opened 2025-03-19
fix: native property of static bytes in puya-ts was throwing unexpected type error

boblat:
## Proposed Changes

- allow `StaticArray<ARC4UIntN<8>, N>` to be passed to ARC4Decode operation to fix the following error when calling `new StaticBytes<32>().native` in `puya-ts`.
  ```
  error: ARC4Decode.value: expression of WType arc4.static_array<arc4.uint8, 33> received, expected arc4.bool or ARC4UIntN or ARC4Tuple or ARC4Struct or ARC4DynamicArray
  ```

### DIFF
--- a/src/puya/ir/builder/arc4.py
+++ b/src/puya/ir/builder/arc4.py
@@ -91,6 +91,11 @@ def decode_arc4_value(
                 args=[value],
                 source_location=loc,
             )
+        case wtypes.ARC4StaticArray(element_type=wtypes.ARC4UIntN(n=8)), (
+            wtypes.bytes_wtype
+            | wtypes.string_wtype
+        ):
+            return value
         case (
             wtypes.ARC4Tuple()
             | wtypes.ARC4Struct() as arc4_tuple,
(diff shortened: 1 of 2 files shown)


src/puya/ir/builder/arc4.py
```diff
@@ -81,10 +81,9 @@ def decode_arc4_value(
                 source_location=loc,
                 types=(PrimitiveIRType.bool,),
             )
-        case wtypes.ARC4DynamicArray(element_type=wtypes.ARC4UIntN(n=8)), (
-            wtypes.bytes_wtype
-            | wtypes.string_wtype
-        ):
+        case wtypes.ARC4DynamicArray(element_type=wtypes.ARC4UIntN(n=8)) | wtypes.ARC4StaticArray(
```
daniel-makerx, line 84: StaticArray's don't have a u16 length header, so this would be incorrectly removing the first 2 bytes of a static array

daniel-makerx: Is it possible to replicate the issue in python so we can have a test case for it?

boblat:
The actual bug was in puya-ts.

PR #432, opened 2025-03-27
Expand ABI method reference type support

daniel-makerx:
Currently the types `Account`, `Application` and `Asset` are only supported as ABI method arguments when used on their own, in these scenarios the values are represented by their ARC-4 aliases and are read from the foreign arrays by the ARC-4 router.

This PR allows their usage in ABI methods when part of native tuples, immutable arrays or on their own as a return type. In these scenarios the values are encoded as their equivalent ARC-4 types. `uint64` for `Asset` and `Application`, and `address` for `Account`

### DIFF
--- a/src/puya/errors.py
+++ b/src/puya/errors.py
@@ -38,21 +38,12 @@ def log_exceptions(fallback_location: SourceLocation | None = None) -> Iterator[
         for code_error in code_errors.exceptions:
             assert isinstance(code_error, CodeError)
             logger.error(code_error.msg, location=code_error.location or fallback_location)  # noqa: TRY400
-    except* InternalError as internal_errors:
-        for internal_error in internal_errors.exceptions:
-            assert isinstance(internal_error, InternalError)
-            _log_traceback()
-            logger.critical(
-                internal_error.msg, location=internal_error.location or fallback_location
-            )
-            sys.exit(ErrorExitCode.internal)
     except* Exception as ex_group:
         for ex in ex_group.exceptions:
-            _log_traceback()
-            logger.critical(f"{type(ex).__name__}: {ex}", location=fallback_location)
-            sys.exit(ErrorExitCode.internal)
-
-
-def _log_traceback() -> None:
-    traceback_lines = traceback.format_exc()
-    logger.debug(traceback_lines.rstrip("\n"))
+            traceback_lines = "".join(traceback.format_exception(ex))
+            logger.debug(traceback_lines.rstrip("\n"))
+            if isinstance(ex, InternalError):
+                logger.critical(ex.msg, location=ex.location or fallback_location)
+            else:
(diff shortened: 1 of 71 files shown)


src/puya/errors.py
```diff
@@ -34,16 +34,23 @@ class CodeError(PuyaError):
...
+            assert isinstance(code_error, CodeError)
+            logger.error(code_error.msg, location=code_error.location or fallback_location)  # noqa: TRY400
+    except* InternalError as internal_errors:
+        for internal_error in internal_errors.exceptions:
+            assert isinstance(internal_error, InternalError)
+            _log_traceback()
+            logger.critical(
+                internal_error.msg, location=internal_error.location or fallback_location
+            )
+            sys.exit(ErrorExitCode.internal)
```
achidlow, line 48: Probably want to log all the errors before exiting..?

src/puya/errors.py
```diff
@@ -34,16 +34,23 @@ class CodeError(PuyaError):
...
+    except* Exception as ex_group:
+        for ex in ex_group.exceptions:
+            _log_traceback()
+            logger.critical(f"{type(ex).__name__}: {ex}", location=fallback_location)
+            sys.exit(ErrorExitCode.internal)
```
achidlow, line 53: Same here?

src/puya/errors.py
```diff
@@ -34,16 +34,23 @@ class CodeError(PuyaError):
...
-        logger.critical(f"{type(ex).__name__}: {ex}", location=fallback_location)
-        sys.exit(ErrorExitCode.internal)
+    except* CodeError as code_errors:
+        for code_error in code_errors.exceptions:
+            assert isinstance(code_error, CodeError)
+            logger.error(code_error.msg, location=code_error.location or fallback_location)  # noqa: TRY400
+    except* InternalError as internal_errors:
+        for internal_error in internal_errors.exceptions:
+            assert isinstance(internal_error, InternalError)
+            _log_traceback()
```
achidlow, line 44: Given this doesn't take any parameters, would this make sense to call outside the loop instead?

src/puya/ir/_contract_metadata.py
```diff
@@ -227,28 +231,55 @@ def _extract_arc4_methods_and_type_refs(
...
+@attrs.frozen
+class _ARC4TypeMapper:
+    errors: list[CodeError] = attrs.field(factory=list)
+
+    @classmethod
+    @contextlib.contextmanager
+    def collect_errors(cls) -> Iterator[typing.Self]:
+        collector = cls()
+        yield collector
+        if collector.errors:
```
achidlow, line 270: If another `CodeError` is raised by something other than through the `.map(...)` function of this object, then any errors collected will be lost as this statement won't execute.

achidlow (approved):

PR #440, opened 2025-04-04
refactor: convert generate_ts_nodes.py to use introspection rather than parsing with MyPy

achidlow:


scripts/generate_ts_nodes.py
```diff
@@ -291,78 +299,51 @@ def print_types(ts_types: list[TsType]) -> Iterable[str]:
...
+            yield from t.render()
+        else:
+            ts_types.append(t)
+            if t.visitor_type and not t.is_abstract:
+                visitor_to_visitable[t.visitor_type].append(t)
+            if awst_nodes.RootNode.__name__ in t.base_types:
+                root_node_types.append(t)
+            yield from t.render()
+
+    # TODO: could this just use RootNode?
```
tristanmenzel, line 319: There's value in having a union type as you can do exhaustive checks

tristanmenzel (approved):

scripts/generate_ts_nodes.py
achidlow: Cool, comment removed.

PR #441, opened 2025-04-08
docs: lsp mermaid diagrams

aorumbayev:
Restoring Neil's https://github.com/algorandfoundation/puya/pull/429 that got lost during rebases on lsp branch. To make it easier, i'm opening it against main branch so that its not affected by ongoing work in lsp branches. 

Additionally added 2 extra flow charts with more details on how lsp works in the python vs typescript frontends


docs/architecture/language-server.md
```diff
@@ -0,0 +1,133 @@
...
+flowchart TB
+    subgraph "VSCode IDE"
+        Developer["Developer"]
+        VSCodeExt["VSCode Extension"]
+    end
+    
+    subgraph "Python Process"
+        PuyaPyLSP["PuyaPy Language Server (puyapy.lsp)"]
+        
+        subgraph "PuyaPy Frontend"
```
neilcampbell, line 44: If i'm not mistaken, both the PuayPy and PuyaTs frontends can also return diagnostics when it fails to produce a valid AWST which will be surfaced by the language server?

neilcampbell (approved):

PR #442, opened 2025-04-14
docs: use scriv to manage change log

boblat:
## Proposed Changes

- check a scriv fragment is added in the pull request
- stop semantic release from generating change log from commit messages
- update contributing.md file to include details on generating scriv fragment 
- update cd pipeline to collect scriv fragments into changelog.md file


achidlow:
Looks good to me, nice and simple - did you find a way to test this out?

boblat:
> Looks good to me, nice and simple - did you find a way to test this out?

Thanks. I was only able to test the individual pieces. Not end to end.

achidlow:
Do you think it's worth testing it out in a fork before we merge? It does require a bit of set up and disabling certain parts (e.g. publishing to PyPI) which is the downside, but on the other hand this pipeline has read/write access to `main`.

.github/workflows/check-python.yaml
```diff
@@ -4,6 +4,22 @@ on:
   pull_request:
 
 jobs:
+  check-change-log-fragment:
```
daniel-makerx, line 7: I like having this check because I am likely to forget to do this at times, however I wonder if it should only be required for feat/fix PR's as sometimes we have PR's that are purely to do with build/chore type tasks.

@achidlow Thoughts?

pyproject.toml
```diff
@@ -32,6 +32,7 @@ colorama = { version = "^0.4.6", markers = "sys_platform == 'win32'" }
 mypy_extensions = "^1.0.0" # vendored mypy dependency
 cattrs = "^24.1"
 networkx = "^3.4.2"
+scriv = "^1.6.2"
```
daniel-makerx, line 35: I think this should be a dev dependency

.github/workflows/check-python.yaml
achidlow: Yeah we can leave it off the required list once it's merged so it just shows up as a check without being required to pass for a merge.

pyproject.toml
boblat: It is now a dev dependency.

boblat:
> Do you think it's worth testing it out in a fork before we merge? It does require a bit of set up and disabling certain parts (e.g. publishing to PyPI) which is the downside, but on the other hand this pipeline has read/write access to `main`.

I have tested it in my forked repo. This is the tag commit semantic-release made as part of CD.
https://github.com/boblat/puya/commit/4dfd910d1388eb6222207aaf0b100ab62cb5ee66

achidlow (approved):

.github/workflows/check-python.yaml
boblat: I am happy with what Adam said.

PR #447, opened 2025-04-17
fix: compile error for using `True`, `False` as default_args value

boblat:
## Proposed Changes

-
-
-

### DIFF
--- a/src/puyapy/awst_build/arc4_decorators.py
+++ b/src/puyapy/awst_build/arc4_decorators.py
@@ -231,6 +231,8 @@ class _ARC4DecoratorArgEvaluator(mypy.visitor.NodeVisitor[object]):
         elif self.arg_name == _CLIENT_DEFAULTS:
             if isinstance(o.node, mypy.nodes.Decorator):
                 return o.name  # assume abimethod
+            if o.fullname in ("builtins.True", "builtins.False"):
+                return _parse_expression(self.context, o)
         return self._resolve_constant_reference(o)
 
     @typing.override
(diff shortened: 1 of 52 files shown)


src/puyapy/awst_build/arc4_decorators.py
```diff
@@ -231,6 +231,8 @@ def visit_name_expr(self, o: mypy.nodes.NameExpr) -> object:
         elif self.arg_name == _CLIENT_DEFAULTS:
             if isinstance(o.node, mypy.nodes.Decorator):
                 return o.name  # assume abimethod
+            if o.fullname in (["builtins.True", "builtins.False"]):
```
achidlow, line 234: lint: I'm surprised ruff didn't flag this, but the () here is redundant (and in this case a bit confusing because it looks like a list within a tuple until you see the lack of comma).

src/puyapy/awst_build/arc4_decorators.py
```diff
@@ -231,6 +231,8 @@ def visit_name_expr(self, o: mypy.nodes.NameExpr) -> object:
         elif self.arg_name == _CLIENT_DEFAULTS:
             if isinstance(o.node, mypy.nodes.Decorator):
                 return o.name  # assume abimethod
+            if o.fullname in ["builtins.True", "builtins.False"]:
```
achidlow, line 234:
```suggestion
            if o.fullname in ("builtins.True", "builtins.False"):
```

Usually we'd use a `tuple` literal here

boblat: I've updated and pushed it.

PR #454, opened 2025-05-01
Add length to bytes wtype

tristanmenzel:
## Proposed Changes

- Add length to bytes wtype


boblat (dismissed):

src/puya/awst/nodes.py
```diff
@@ -1368,13 +1372,13 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 @attrs.frozen
 class BytesBinaryOperation(Expression):
```
achidlow, line 1373: The changes to this class mean the size of the result type isn't validated - and by default, it will be incorrect.

src/puya/awst/wtypes.py
```diff
@@ -29,6 +29,19 @@ def __str__(self) -> str:
         return self.name
 
 
+@attrs.frozen(kw_only=True)
+class BytesWType(WType):
+    length: int | None = attrs.field(eq=False)
```
achidlow, line 34: Size shouldn't be excluded from equality checks, especially if we're relying on it for correctness (e.g. with the loop code change you added).

tristanmenzel: This is different to what we discussed the other day but shouldn't be a problem

src/puya/awst/nodes.py
tristanmenzel: I have refactored the node to use a validator instead of a default value which checks the expression type is valid for the operands and operator

src/puya/awst/nodes.py
```diff
@@ -1372,7 +1372,11 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 class BytesUnaryOperation(Expression):
     op: BytesUnaryOperator
```
tristanmenzel, line 1373: Add validator in [BytesUnaryOperator.bitInvert]

src/puya/awst/nodes.py
```diff
@@ -405,8 +405,8 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 
 @attrs.frozen
 class MethodConstant(Expression):
-    wtype: WType = attrs.field(default=wtypes.bytes_wtype, init=False)
     value: str
+    wtype: WType = attrs.field(default=wtypes.bytes_wtype)
```
achidlow, line 409: This needs validation that it's a bytes type - and either unsized or `length=4`

src/puya/awst/wtypes.py
achidlow: Ah I recall the conversation more clearly now! Yes, that was under the assumption that the only place this information would end up is in the ARC-56. So if we did that and removed the lowering, then yeah, we could document that this field is only a hint, and then we ignore any sizing when doing both AWST validation and IR lowering.

achidlow:
@tristanmenzel just pushed up some commits that I think that finishes off all the missing validation.

src/puya/awst/nodes.py
achidlow: Did this with a static type assert instead

daniel-makerx:
In general looks good, but need a test case to ensure the backend is exercising this new code

daniel-makerx (approved):

PR #456, opened 2025-05-05
docs: document to discuss proposed PuyaTS 1.0 changes

daniel-makerx:


docs/architecture-decisions/2025-05-01_puyats-1.0.md
```diff
@@ -179,16 +179,15 @@ return foo; // { a: 3, b: 2 }
 
 ### PuyaTS 1.0
 ```typescript
-type Foo = {
```
joe-p: Does this change mean objects can not be mutated in 1.0?

tristanmenzel: The decision is not so much about what a class gives us over a pojo, it’s about what the type system will let us differentiate. 

On that note, I’m currently spiking a solution where we use Readonly<{…}> to declare an immutable object and objects are mutable by default. It does mean less optimal (mutable) types will be used by default and the Readonly type will have to be added to optimise. Perhaps this is ok?

neilcampbell (approved):

PR #458, opened 2025-05-06
feat(internal): Add SetInnerTransactionFields node 

tristanmenzel:
## Proposed Changes

- Add SetInnerTransactionFields node which abstracts em…ission of `itxn_field` ops for a sequence of WinnerTransactionFields expressions delimited with `itxn_begin` or `itxn_next` ops.

### DIFF
--- a/src/puya/ir/builder/itxn.py
+++ b/src/puya/ir/builder/itxn.py
@@ -19,6 +19,7 @@ from puya.ir.builder._utils import assign, assign_intrinsic_op
 from puya.ir.builder.blocks import BlocksBuilder
 from puya.ir.context import IRFunctionBuildContext
 from puya.ir.models import (
+    BasicBlock,
     ConditionalBranch,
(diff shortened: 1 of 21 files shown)


src/puya/ir/builder/itxn.py
```diff
@@ -653,6 +678,13 @@ def __init__(self) -> None:
         # used to ensure multiple submits are not present
         self.seen_submit_exprs = set[awst_nodes.SubmitInnerTransaction]()
 
+    @typing.override
+    def visit_set_inner_transaction_fields(
+        self, _node: awst_nodes.SetInnerTransactionFields
+    ) -> list[_SourceAction]:
+        # TODO: Confirm nothing is required here
```
daniel-makerx, line 685: Can move to the idc region and use `return self._empty_actions_from_wtype(expr)`

boblat (approved):

PR #468, opened 2025-05-12
Create dependabot.yml

lempira:
Adding dependabot for package version updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,9 @@
 
 version: 2
 updates:
-  - package-ecosystem: "pip"
-    directory: "/"
-    schedule:
-      interval: "weekly"
+    - package-ecosystem: "pip"
+      directory: "/"
+      schedule:
+          interval: "weekly"
+      ignore:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
neilcampbell, line 11:
```suggestion
      interval: "weekly"
      ignore:
        - dependency-name: "mypy"
```

The mypy version needs to be carefully managed, so it's worth ignoring.

lempira: done

neilcampbell (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+    - package-ecosystem: "pip"
```
achidlow, line 8: The GitHub docs say [this only supports Poetry v1](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-), but we're on v2?

lempira: I noticed this as well when I looked at the list. We are using poetry v2 in other py repos and this seems to be working. I think what's happening is that we are not specifying a min requirement for poetry (the dep management tool) in the pyproject file. We are only specifying the min version of `poetry-core` for building. Because of that, I think GH is using it's version of poetry for dependabot.

achidlow: Ah interesting, well let's give a try then and see how it goes!

achidlow (approved):

daniel-makerx (approved):

PR #497, opened 2025-07-17
docs: add ADR for reference aliasing change

daniel-makerx:


joe-p:
Needing to explicitly use references is probably such a rare thing developers need to do that I don't think it really matters how we do it. I would say the main consideration is migration. If we require a decorator arg, it will be hard to inform the user that their contract behavior is changing (like we could with `@deprecated` for example). This will be covered in the migration guide and changelog, so maybe that's not as important?

Given the rarity of needing to use it I think the awkwardness with two different types for the same underlying resource would be ok. An alternative is deprecate existing type and introduce a new type that returns the index of the respective array, but that might be over-complicating it. 

TL;DR Don't have a strong opinion since its seldomly used, but we should consider migration path.

daniel-makerx:
It looks like the bigger question is how to best communicate this change to users, some things to consider with this change

* Where users are at
  * Existing users, using reference behaviour
  * Existing users, with a custom solution using id behaviour
  * New users
* What tooling they are using
  * algosdk
  * algokit_utils
  * custom
* Where we want them to be

Different options for implementing this change

1. No change in default behaviour
  * Don't force a change on the user
  * Add the capability for desired behaviour. 
  * Templates and documentation could guide to desired id behaviour
  * Low friction, but reduces uptake of new behaviour
  * Allows a period to communicate and try the change before making it the default
2. Change behaviour silently
  * Force a change on the user but don't require changes to contracts/configuration
  * Outputs would change
  * May not be immediately clear there is a change if they aren't using approvals or reading release notes
  * Templates and documentation would use desired ID behaviour
  * Low friction and increases uptake of new behaviour but may be unexpected
3. Change behaviour with a warning until the user explicitly chooses something
  * Force a change on the user but don't require changes to contracts/configuration
  * Would require them to acknowledge the change via contract/configuration to silence warning
  * Templates and documentation would use desired ID behaviour
  * Low friction, increases uptake of new behaviour, some communication that something has changed
4. Force a change in behaviour with an error, user must pick a path forward
  * Force a change on the user, requires a change to contracts/configuration
  * High friction and forces user to consider implications

Also worth considering these options for both PuyaPy and PuyaTS

PuyaPy will require one of the above
PuyaTS 1.0 could go straight to option 2, since a transition from Beta to 1.0 is a natural transition point

PR #500, opened 2025-07-21
build: add prerelease CD workflow

daniel-makerx:


.github/workflows/prerelease.yaml
```diff
@@ -0,0 +1,114 @@
...
+  contents: write
+  packages: read
+env:
+  SEMANTIC_RELEASE_DRY_RUN: ${{ inputs.dry_run && '--noop' || '' }}
+jobs:
+  prepare-release:
+    runs-on: ubuntu-latest
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
```
achidlow, line 24: Does it work without this since it's not publishing to PyPI?

PR #505, opened 2025-07-22
chore: update mypy to v1.17.0

achidlow:


src/puyapy/parse.py
```diff
@@ -314,7 +314,8 @@ def get_mypy_options(
 
     # disallow use of any
     mypy_opts.disallow_any_unimported = True
-    mypy_opts.disallow_any_expr = True
+    # should be True but: https://github.com/python/mypy/issues/18770
+    mypy_opts.disallow_any_expr = False
```
daniel-makerx, line 318: I assume if a user *did* write an any expression it would still fail to resolve a pytype, so even though there may not be a mypy error it will still fail compilation

daniel-makerx (approved):

Issue #453, opened 2025-04-30
urange assigning loop value at stop index


atsoc1993:
In traditional Python, when you create a for loop to iterate through a range, the last value of the loop variable will never be the stop index:

eg:
```
for i in range(0, 1):
    print(i)
print(i)
```

Output:
```
0
0
```

When using urange, the result is not consistent:
```for i in urange(0, 1):
    log(i)
log(i)
```

Output:
```0
1
```

Below is a simple test contract and deployment code you can copy/paste to quickly confirm:

```from algopy import ARC4Contract, urange, log
from algopy.arc4 import abimethod


class Test(ARC4Contract):
    def __init__(self) -> None:
        pass

    @abimethod
    def log_i_in_urange(self) -> None:
        for i in urange(0, 1):
            log(i)
        log(i)
```

## Retrieve client with `algokitgen-py -a filename.arc56.json -o clientname.py`

```
from algokit_utils import AlgorandClient
from Test_Client import TestFactory
from algosdk.atomic_transaction_composer import AccountTransactionSigner
from algosdk.abi import ABIType
from base64 import b64decode

sk = '' 
address = ''
signer = AccountTransactionSigner(sk)

algorand = AlgorandClient.testnet()

app_client, deploy_response = TestFactory(
    algorand=algorand,
    default_sender=address,
    default_signer=signer,
).send.create.bare()


txn_response = app_client.send.log_i_in_urange()

uint64_coder = ABIType.from_string('uint64')
logs = txn_response.confirmation['logs']

for log in logs:
    print(uint64_coder.decode(b64decode(log)))

```

PR #529, opened 2025-08-18
fix: urange(.) iteration index off by one issue

Argimirodelpozo:
Changed order of for blocks in iterate_urange_simple(.) in order to prevent the iteration variable from reaching the end of the range. This better matches python behavior. See issue #453 .


Argimirodelpozo:
The final solution I ended up going with keeps the general structure of `_urange_simple(.)`, but checks at the end if the iteration variable changed and therefore loop execution happened. In that case, `step` is subtracted so the iteration variable becomes the last valid element of the range. The index is treated the same if necessary.
I found this to be the simplest I could do in terms of adding extra complexity.

achidlow:
Nice work trying to find a simple solution, unfortunately this will produce incorrect results if the user variables are updated inside in the loop:

```python
for i in urange(5):
    i = UInt64(100)
assert i == 100
```

I think the only correct way to address this is to use internal variables for the range item (and range index if enumerating), and assign them to the user variables at the start of the loop body.

PR #532, opened 2025-08-28
Perf improvements

daniel-makerx:


src/puya/ir/optimize/collapse_blocks.py
```diff
@@ -87,69 +24,82 @@ def _replace_single_target_with_goto(terminator: models.ControlOp) -> models.Con
...
+        if len(block.predecessors) == 1:
+            (predecessor,) = block.predecessors
+            if type(predecessor.terminator) is models.Goto:
+                assert predecessor.terminator.target is block
                 # can merge blocks when there is an unconditional jump between them
                 predecessor.phis.extend(block.phis)
                 predecessor.ops.extend(block.ops)
+                predecessor.terminator = block.terminator
+                if predecessor.label is None:
+                    predecessor.label = block.label
```
daniel-makerx, line 41: Do we need to back propagate the label in this case? Might be confusing when trying to understand the IR CFG

PR #537, opened 2025-09-02
ci: add token bot with write perms

daniel-makerx:
Same PR as #494, trying from this repo instead of a fork to see if the secrets are accessible


.github/workflows/check-python.yaml
```diff
@@ -3,10 +3,21 @@ name: Check Python Code
 on:
   pull_request:
 
+permissions:
+  pull-requests: write # needed for thollander/actions-comment-pull-request
+
 jobs:
+
```
achidlow, line 10: 👨‍🚀

achidlow (approved):

PR #538, opened 2025-09-03
Fix/urange iteration variable value outside of loop

Argimirodelpozo:
This is a solution to issue #453. Moved from a fork to a branch in the repo (from this PR #529) to get some CI/CD actions to run.


src/puya/ir/builder/iteration.py
```diff
@@ -282,44 +282,80 @@ def _iterate_urange_simple(
     loop_vars = assigner.assign_user_loop_vars(
```
achidlow, line 282: So this pre-existing code actually presents another small semantic incompatibility, I've realised, and we should fix it as part of this PR.

Because we don't pre-check if we will loop here like in `_iterate_urange_with_reversal`, we shouldn't be assigning any values to user variables until the loop begins. Since if the range is empty, then the "first" value is undefined. This matches Python behaviour and also makes sense logically.

Fortunately, all we need to do here is just delete this line. The internal variables can be directly assigned `start` and a zero constant instead of copying these user variables. Unfortunately, as I tested this out, it will result in a small size increase in some contracts - but just a small one, not anything to concern ourselves with.

test_cases/iteration/iterate_urange.py
```diff
@@ -13,6 +13,7 @@ def test_forwards(self) -> None:
         for i in urange(1, 7, 2):
             log(values[i])
             i += 1
+        assert i == UInt64(6)
```
achidlow, line 16: Whenever there's a binary operation, as long as one operand type has an algopy type, the other operand can be the equivalent literal type. So this can just be `assert i == 6`.

src/puya/ir/builder/iteration.py
Argimirodelpozo: Absolutely. So I noticed this difference when I was doing some preliminary testing of the current behavior (when I tried an assignment after the urange iteration test for empty ranges) but since I was doing aso precheck in my previous attempt anyways I forgot. But since now the first effective assignment is inside the loop body you're totally right, it's right there if we remove that assignment that is unnecesary to begin with.

PR #551, opened 2025-09-08
Partial mypy refactoring

achidlow:
This PR makes incremental progress towards reducing the puyapy dependency on mypy.

It also updates to the latest mypy version.


pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
 [dependency-groups]
 dev = [
     "algorand-python",
-    "mypy==1.17.0",
```
daniel-makerx, line 34: Just double checking mypy didn't introduce any more dependencies in this version?

Issue #498, opened 2025-07-18
Named Tuple argument evaluation order is incorrect


daniel-makerx:
When a NamedTuple is constructed, the arguments are evaluated in the order they are defined on the type, rather than the order they are in when the tuple is instantiated. This is a semantic incompatibility bug, with observable differences if any of the arguments have side effects e.g.

```python
import typing

from algopy import UInt64, arc4, log, subroutine


class MyTuple(typing.NamedTuple):
    one: UInt64
    two: UInt64


class MyContract(arc4.ARC4Contract):
    @arc4.abimethod
    def create_tuple(self) -> MyTuple:
        return MyTuple(two=echo(UInt64(2)), one=echo(UInt64(1)))


@subroutine
def echo(value: UInt64) -> UInt64:
    log(value)
    return value

```

### Expected behaviour
Contract logs `2, 1` and then the ABI return value

### Actual behaviour
Contract logs `1, 2` and then the ABI return value

PR #557, opened 2025-09-14
Fix Named Tuple argument evaluation order

Argimirodelpozo:
This PR addresses issue #498.
Previously, when lowering `NamedTuple` instantiations, the IR ignored the evaluation order of arguments in the source expression and instead always used the field order defined in the type.
To fix this, a new AWST node, `NamedTupleExpression`, has been introduced. Its structure is similar to the existing `NewStruct` node, preserving the original argument evaluation order in its `values` mapping. During lowering, arguments are now evaluated in this order, and then rearranged to match the type definition so that subsequent passes continue to work as before.
Additionally, the example from the issue has been incorporated as a new test case under `test_cases/named_tuples`.


src/puya/awst/nodes.py
```diff
@@ -962,6 +962,29 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
...
+    Defines a new named tuple with the specified elements.
+    Elements are evaluated in mapping order.
+    """
+
+    wtype: wtypes.WTuple
+    values: Mapping[str, Expression] = attrs.field(converter=immutabledict)
+
+    @values.validator
+    def _validate_values(self, _instance: object, values: Mapping[str, Expression]) -> None:
+        if values.keys() != self.wtype.fields.keys():
```
achidlow, line 977: This should probably first validate that `WTuple.names` is not `None` - if it is then that's an `InternalError`.

src/puya/awst/nodes.py
```diff
@@ -962,6 +962,29 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
...
+            raise CodeError("Invalid argument(s)", self.source_location)
```
achidlow, line 978: I can see this uses the same code as NewStruct which is good, however we're in the process of making error messages start with lower case (this is rather inconsistent at the moment, but there are some larger changes planned here around error reporting for the language server, and warning treatments as well generally, so this work is more of a fix-it-as-you-go thing).

src/puya/awst/to_code_visitor.py
```diff
@@ -502,6 +502,11 @@ def visit_tuple_expression(self, expr: nodes.TupleExpression) -> str:
             items = ", ".join([item.accept(self) for item in expr.items])
         return f"({items})"
 
+    @typing.override
+    def visit_named_tuple_expression(self, expr: nodes.NamedTupleExpression) -> str:
+        items = ", ".join([f"{name}={value.accept(self)}" for name, value in expr.values.items()])
```
achidlow, line 507: The list comprehension is technically redundant here, although I can see it's come from copying the behaviour for NewStruct also.

src/puyapy/awst_build/eb/tuple.py
```diff
@@ -133,14 +134,17 @@ def call(
...
-        # TODO: need to evaluate values args in field_mapping order
-        values = [
-            expect.argument_of_type_else_dummy(field_mapping[field_name], field_type).resolve()
-            for field_name, field_type in pytype.fields.items()
-        ]
-        expr = TupleExpression(
-            items=values,
+        expr = NamedTupleExpression(
             wtype=pytype.wtype,
+            values=[
```
achidlow, line 139: This works because of the attrs field converter, but to be closer to the true type you can just use a dictionary comprehension here. 

Also I'd suggest having the dictionary comprehension as it's own statement for readability.

achidlow:
Looking good, just some minor changes.
Also we try to avoid merge commits generally, and maintain a linear history.
I think for this PR ideally it would be 2-4 commits. One with the feature and changelog and maybe Python changes to test contracts (this could go in it's own PR if preferred), and one or two compile all commits (either just one with all the changes, or one after the fix but before the additional tests and one after the additional tests).

PR #590, opened 2025-10-27
feat: add fixed size variant of bytes as a separate `FixedBytes` type

boblat:


stubs/algopy-stubs/_primitives.pyi
```diff
@@ -299,6 +311,86 @@ class BytesBacked(typing.Protocol):
     def bytes(self) -> Bytes:
         """Get the underlying Bytes"""
 
+_TBytesLength = typing.TypeVar("_TBytesLength", bound=int)
+
+class FixedBytes(typing.Generic[_TBytesLength], BytesBacked, _BytesBase):
```
daniel-makerx: We should make FixedBytes a validatable type as well

src/puyapy/awst_build/pytypes.py
```diff
@@ -570,6 +570,40 @@ def wtype(self) -> ErrorMessage:
...
+    try:
+        (length_t,) = args
+    except ValueError:
+        raise CodeError(
+            f"expected a single type parameter, got {len(args)} parameters", source_location
+        ) from None
+    length = _require_int_literal(self, length_t, source_location)
+    if length < 0:
+        raise CodeError("FixedBytes length should be non-negative", source_location)
+    if length > algo_constants.MAX_BYTES_LENGTH:
```
daniel-makerx: Boxes can contain more bytes than the max bytes that can fit on the stack, so we should not constrain the upper limit

test_cases/fixed_bytes_ops/abi-values.py
daniel-makerx: python module name should be `abi_values.py`

src/puyapy/awst_build/pytypes.py
achidlow: Ah, good point, I think Bobby did have this originally elsewhere and I moved it here, whoops

stubs/algopy-stubs/_primitives.pyi
boblat: it is now a validatable type now

src/puyapy/awst_build/eb/fixed_bytes.py
```diff
@@ -0,0 +1,630 @@
...
+            return dummy_statement(location)
+
+        target = self.resolve_lvalue()
+        match rhs:
+            case InstanceBuilder(pytype=pytypes.BytesLiteralType):
+                other = rhs.resolve_literal(FixedBytesTypeBuilder(self.pytype, location)).resolve()
+            case InstanceBuilder(pytype=ptype) if isinstance(ptype, pytypes.FixedBytesType):
+                other = rhs.resolve()
+            case InstanceBuilder(pytype=pytypes.BytesType):
+                # perform runtime length check and cast to FixedBytes
```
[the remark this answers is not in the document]
boblat: guard and test have been added for `+` operator

achidlow (approved):

PR #599, opened 2025-11-06
refactor: add @public as alias of @abimethod and relax requirement of @subroutine

boblat:


stubs/algopy-stubs/__init__.pyi
```diff
@@ -30,3 +30,6 @@ from algopy.op import (
     Global as Global,
     Txn as Txn,
 )
+
+# public decorator is an alias for abimethod to provide a cleaner API
+public = arc4.abimethod
```
achidlow: This will need a stubs version bump.

daniel-makerx: It was already bumped for [fixed bytes](https://github.com/algorandfoundation/puya/pull/590/files#diff-dc513b523cabfe1686c4575cb4748b03c5f981e1e839b5fe424caef742027c28L7) which is unreleased, so another bump isn't necessary?

All tho worth noting the fixed bytes bump was incorrect, it should have been `3.2.0` not `3.2.1` So could fix that in this PR, or have another PR to fix that

achidlow (approved):

PR #602, opened 2025-11-13
feat: improved sized bytes code generation

achidlow:
Handle encoding/decoding sized native bytes to ARC-4 arrays more efficiently, and also optimize len of extract with known size.


src/puya/ir/builder/aggregates/arc4_codecs.py
```diff
@@ -493,6 +493,58 @@ def decode(
         return None
 
 
+@attrs.frozen
+class _SizedBytesCodec(_ScalarCodec):
+    length: int
+    element: encodings.UIntEncoding | encodings.UTF8Encoding
```
daniel-makerx: don't need UTF8 encoding at present?

daniel-makerx (approved):

PR #607, opened 2025-11-27
feat: add `stage` and `submit_staged` functions to support submitting dynamic number of inner transactions in a group

boblat:


src/puya/awst/function_traverser.py
```diff
@@ -168,6 +168,11 @@ def visit_update_inner_transaction(self, call: awst_nodes.UpdateInnerTransaction
         for value in call.fields.values():
             value.accept(self)
 
+    @typing.override
+    def visit_stage_inner_transactions(self, node: awst_nodes.StageInnerTransactions) -> None:
+        for expr in node.itxns:
```
achidlow: should visit `node.start_new_group` also since it's an `Expression`

src/puya/awst/to_code_visitor.py
```diff
@@ -472,6 +472,13 @@ def visit_update_inner_transaction(self, expr: nodes.UpdateInnerTransaction) ->
             fields.append(f"{field.immediate}={value.accept(self)}")
         return f"update_inner_transaction({expr.itxn.accept(self)},{', '.join(fields)})"
 
+    @typing.override
+    def visit_stage_inner_transactions(self, node: nodes.StageInnerTransactions) -> str:
+        start_new_group = node.start_new_group.accept(self)
+        itxns = f'{", ".join(itxn.accept(self) for itxn in node.itxns)}'
+
+        return f"stage(itxns=[{itxns}], start_new_group={start_new_group})"
```
achidlow: Not that it matters too much, but maybe we should call this `stage_itxns` instead of having the `itxns=` bit?
Just because otherwise it might look like a subroutine name. I mean, either way, this output here is mostly for debugging the front end so it doesn't matter too much.

stubs/algopy-stubs/itxn.pyi
```diff
@@ -61,6 +61,9 @@ _TResult_co = typing.TypeVar(
 )
 
 class _InnerTransaction(typing.Protocol[_TResult_co]):
+    def stage(self, *, begin_group: bool = False) -> None:
```
achidlow: All good as things currently stand, but if we do a release before this is merged, will need to bump stubs version again.

achidlow: compiler v5.5.0 + stubs v3.2.0 has been released so will need to bump stubs after a rebase

achidlow (approved):

PR #615, opened 2025-12-19
refactor: push down type mapping logic for arc4_signature into IR layer

boblat:


test_cases/abi_routing/out/module.awst
```diff
@@ -223,7 +223,7 @@ contract Reference
     GlobalState['creator']: account = txn<Sender>()
     GlobalState['app']: application = reinterpret_cast<application>(123u)
     assert(len(Method("get(uint64,byte[])byte[]")) != 0u, comment="has method selector")
-    assert(Method("opt_in(uint64,byte[])void") == Method("opt_in(uint64,byte[])void"), comment="has method selector for method reference")
+    assert(Method(signature="opt_in(uint64,bytes)void", resource_encoding="value") == Method("opt_in(uint64,byte[])void"), comment="has method selector for method reference")
```
daniel-makerx: Minor: It might be worth changing `to_code_visitor.py` to make the `name`, `args` and `result` values separate arguments, as the signature here seems like it should be an ARC-4 signature, when in fact it is using wtype names, which can be different (`bytes` vs `byte[]`)

src/puya/ir/types_.py
```diff
@@ -453,3 +454,120 @@ def ir_type_to_ir_types(ir_type: IRType | TupleIRType) -> list[IRType]:
...
+    def __init__(
+        self,
+        loc: SourceLocation | None,
+        resource_encoding: typing.Literal["index", "value"] = "value",
+    ) -> None:
+        self.loc = loc
+        self.resource_encoding = resource_encoding
+
+    def visit_basic_type(self, wtype: wtypes.WType) -> str:
+        if wtype == wtypes.biguint_wtype:
```
daniel-makerx: For these basic types might be nicer as a mapping of wtype to arc4 name? Probably also need a second mapping before that to first map the reference types when resource_encoding is value

src/puya/ir/arc4_router.py
```diff
@@ -530,7 +530,10 @@ def _create_abi_switch(
     for method, method_wrapper in cases:
         abi_loc = method.config_location
         method_const = awst_nodes.MethodConstant(
-            source_location=router_location, value=method.signature
+            source_location=router_location,
```
daniel-makerx: Minor: normally try to order kwargs in order or significance, so generally source_location is last

src/puya/awst/nodes.py
```diff
@@ -542,6 +543,31 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
...
+class MethodSignature(Node):
+    name: str = attrs.field()
+    arg_types: Sequence[wtypes.WType] = attrs.field(converter=tuple[wtypes.WType, ...])
+    return_type: wtypes.WType
+    resource_encoding: typing.Literal["index", "value"] = "value"
+
+    @name.validator
+    def _name_validator(self, _attr: object, value: str) -> None:
+        if not self.is_valid_method_name(value):
+            raise CodeError(f"method name '{value}' does not conform to ARC-4 spec")
```
daniel-makerx: We have this as a warning in AWST validation, I wonder if should be a warning here too?

PR #622, opened 2025-12-29
Migrates puya testing infrastructure to use algokit utils v5

daniel-makerx:
### Notable changes

Tests in `tests/test_arc32.py` and `tests/test_execution.py` have all been migrated to individual test suites within `tests/onchain`

`trace.log` files are no longer produced. These provide limited benefit as approvals and are superseded by simulation traces if execution traces are required in the future.

Removed unused dryrun script.


neilcampbell:
Changed to draft for now, since we need to have a conversation to decide if we're ok to merge now.

tests/onchain/test_amm.py
```diff
@@ -0,0 +1,159 @@
...
+
+        return {
+            h.asset_id: h.amount
+            for h in account_info.assets or ()
+            if h.asset_id in (pool_token, asset_a, asset_b)
+        }
+
+    def get_ratio() -> int:
+        app_client.get_global_state()
+        state = app_client.get_global_state()
```
PatrickDinh: I think the first `app_client.get_global_state()` is not needed

tests/from_awst/itxn_compose/test_itxn_compose.py
```diff
@@ -13,66 +10,71 @@
...
-    pay = TransactionWithSigner(
-        txn=algosdk.transaction.PaymentTxn(
-            sender=account.address,
+    addresses = []
+    for _ in range(3):
+        a = localnet.account.random()
+        localnet.account.ensure_funded(
+            account_to_fund=a.addr,
+            dispenser_account=account,
+            min_spending_balance=au.AlgoAmount.from_algo(1),
```
Argimirodelpozo: I think there is a discrepancy in the algo amounts? (1000 vs. MBR+1). Don't think it matters for the test anyways.

Argimirodelpozo (commented):
Just finished reviewing. Looks good to merge!

tests/from_awst/itxn_compose/test_itxn_compose.py
daniel-makerx: Yeah 1000 seemed excessive for these accounts, I think it was only chosen because it was the old functions default value. Since the tests didn't need that much I dropped it to just be 1 algo of spendable funds.

PR #624, opened 2026-01-06
feat: add new itxn.abi_call method to allow calls to ABI method be part of dynamic itxn group

boblat:


daniel-makerx (approved):

Argimirodelpozo (commented):
[edit: late night misclick]
It looks good to merge!

src/puya/ir/builder/aggregates/arc4_codecs.py
```diff
@@ -346,6 +346,8 @@ def encode_value(
     ) -> ir.ValueProvider | None:
         match encoding:
             case encodings.UIntEncoding(n=bits):
+                if isinstance(value, ir.UInt64Constant) and value.value >= (1 << bits):
```
Argimirodelpozo: Probably missing something but why don't we raise an error or assert here?

PR #625, opened 2026-01-12
Small refactor in TEAL optimizations

Argimirodelpozo:
While writing specs. for this part noticed that there was a redundant if condition, and that removal of fallthrough branches for `O0` was unreachable.
This seemed to be unintentional as the `_remove_jump_fallthroughs(.)` function does hanlde the special case where the branch has stack manipulations and a comment implying it should be run under `O0` (which makes sense to me).
If I'm missing something here please do let me know.
Thanks!

### DIFF
--- a/src/puya/teal/optimize/main.py
+++ b/src/puya/teal/optimize/main.py
@@ -24,6 +24,9 @@ logger = log.get_logger(__name__)
 def optimize_teal_program(
     context: ArtifactCompileContext, teal_program: models.TealProgram
 ) -> None:
+    # Note: at optimization level O0, this function still applies the minimal
+    # transformations required to ensure the program is valid and functional
+
     # O0 will still remove some redundant ops to ensure a feasible program size
     for teal_sub in teal_program.all_subroutines:
         _optimize_subroutine_ops(context, teal_sub)
@@ -40,6 +43,8 @@ def optimize_teal_program(
             _optimize_subroutine_blocks(context, teal_sub, branchable_subroutine_entry_blocks)
         maybe_output_intermediate_teal(context, teal_program, qualifier="block")
 
+    # O0 still needs to gather template vars into constant blocks
+    # to compute pc offset according to arc56
     gather_program_constants(teal_program)
     if context.options.optimization_level > 0:
         combine_pushes(teal_program)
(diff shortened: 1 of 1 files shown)


src/puya/teal/optimize/main.py
```diff
@@ -61,23 +65,22 @@ def _optimize_subroutine_blocks(
         f"optimizing TEAL subroutine blocks {teal_sub.signature}",
         location=teal_sub.source_location,
     )
-    if context.options.optimization_level > 0:
```
daniel-makerx: Good spot, some additional conditions were added to remove optimizations at O0 in order to speed up the language server analysis, which then resulted in this redundant check

src/puya/teal/optimize/main.py
```diff
@@ -39,6 +39,10 @@ def optimize_teal_program(
         for teal_sub in teal_program.all_subroutines:
             _optimize_subroutine_blocks(context, teal_sub, branchable_subroutine_entry_blocks)
         maybe_output_intermediate_teal(context, teal_program, qualifier="block")
+    else:
```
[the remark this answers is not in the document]
Argimirodelpozo: Oooooh ok, makes sense. It just threw me off that the function has a comment implying that it should be run under O0. I'll try to add the comment you suggested and maybe slightly reword the one in `_remove_jump_fallthrough(.)`

src/puya/teal/optimize/main.py
```diff
@@ -39,6 +39,10 @@ def optimize_teal_program(
...
+        # jump fallthroughs should still be removed in O0
+        for teal_sub in teal_program.all_subroutines:
+            _remove_jump_fallthroughs(teal_sub)
 
     gather_program_constants(teal_program)
```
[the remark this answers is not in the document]
Argimirodelpozo: Thanks! Will add a comment on the spec. section for this.

src/puya/teal/optimize/main.py
```diff
@@ -283,14 +287,14 @@ def _remove_jump_fallthroughs(teal_sub: models.TealSubroutine) -> None:
     for block, next_block in zip(teal_sub.blocks, teal_sub.blocks[1:], strict=False):
         match block.ops[-1]:
             # we guard against having stack manipulations but only one op, thus nowhere to put
-            # them, but this should only occur in O0 as in higher levels, blocks with just a b
-            # will already be inlined
+            # them, but this should only occur in O0 (which does not reach this function)
+            # in higher levels, blocks with just a b will already be inlined
```
achidlow:
```suggestion
            # If there are stack manipulations, we require somewhere to put them within the same
            # block, ie at least one remaining op.
            # In theory, we could change the way we store stack manipulations to allow attaching
            # them to block entry/exit, but in practice  this optimisation is only running at O1
            # and above, and currently that means inlining of jump chains also occurs (and thus in
            # theory there should not be any blocks with just a branch here), making such a change
            # redundant.
```

src/puya/teal/optimize/main.py
```diff
@@ -283,14 +287,14 @@ def _remove_jump_fallthroughs(teal_sub: models.TealSubroutine) -> None:
...
             case models.Branch(
                 target=target_label, stack_manipulations=sm
             ) if target_label == next_block.label and (len(block.ops) > 1 or not sm):
                 logger.debug(f"removing explicit jump to fall-through block {next_block.label}")
                 block.ops.pop()
-                if block.ops:  # guard is only for O0 case
+                if block.ops:  # guard is an extra precaution as this should only apply to O0
```
achidlow:
```suggestion
                if block.ops:
```

PR #645, opened 2026-02-03
feat: add support for boxes > 4k containing dynamic arrays of fixed elements

daniel-makerx:


Argimirodelpozo (commented):
Looks great!

src/_puya_lib/box_arc4.py
```diff
@@ -0,0 +1,106 @@
...
+    # update length header of array
+    arr_len = _box_extract_u16(box_key, array_offset)
+    new_arr_len = arr_len - 1  # on error: empty array
+    index = new_arr_len  # TODO: support popping by index by making this an argument
+    new_arr_len_u16 = _as_uint16(new_arr_len)
+    op.Box.replace(box_key, array_offset, new_arr_len_u16)
+
+    # remove element from array
+    popped_item_offset = array_offset + 2 + index * fixed_byte_size
+    op.Box.splice(box_key, popped_item_offset, fixed_byte_size, b"")
```
Argimirodelpozo: Wait. Could we just resize here since we are erasing the last element?

Argimirodelpozo: I guess this is leaving room for popping by index, but maybe we could have an extra function for that and save the _one_ measly box op whenever we know its the last element we want gone :p

daniel-makerx: Additionally the array may not be the last thing in the box, for example if there is a struct with two dynamic arrays the tail portion of both are at the end of the struct

src/_puya_lib/box_arc4.py
```diff
@@ -0,0 +1,105 @@
...
+    new_items_count: UInt64,
+    fixed_element_size: UInt64,
+) -> None:
+    """
+    Modifies a box's content by concatenating data to an arc4 dynamic array of fixed size elements
+
+    box_key: The box_key to manipulate
+    array_offset: The offset in bytes to the start of the array
+    new_items_count: The count of new items being added (N)
+    new_items_bytes: THe concatenated bytes of N array elements
```
achidlow: nit: `THe`

PR #648, opened 2026-02-05
Improve circular import handling

achidlow:
Currently, we order the Strongly Connected Components (SCC) in PuyaPy by name - we should be sorting them by dependency type. e.g., if A imports B and B imports A, but the former is in a TYPE_CHECKING block, we should process B first.

Also fixes an issue with source locations that arose in the testing of this.

### DIFF
--- a/changelog.d/20260205_101919_achidlow_scc_ordering.md
+++ b/changelog.d/20260205_101919_achidlow_scc_ordering.md
@@ -1,0 +1,42 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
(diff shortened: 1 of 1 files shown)


test_cases/circular_dependency/b.py
```diff
@@ -19,3 +36,8 @@ def create_a(self, factory_id: Application) -> Application:
 
         result, _txn = arc4.abi_call(Factory.deploy_a, app_id=factory_id)
         return result
+
+
+@subroutine
+def check_inner(s: "MyInnerStruct") -> None:
```
daniel-makerx: Should this test case include calling this function?

PR #651, opened 2026-02-09
refactor: move mapping to arc-4 type down to ir layer for ARC4Encode and ARC4Decode 

boblat:


daniel-makerx:
> > ❌ Stubs version must be bumped using `poe bump_stubs` when .pyi files in stubs directory are changed.
> 
> Only doc string for `size_of` method is changed to mention it works only with statically sized types

Should still bump for this scenario, but can do a patch bump e.g. `poe bump_stubs 3.4.1`

src/puya/ir/builder/main.py
```diff
@@ -150,14 +150,17 @@ def visit_arc4_encode(self, expr: awst_nodes.ARC4Encode) -> TExpression:
...
+        else:
+            encoded_ir_type = types.wtype_to_encoded_ir_type(expr.wtype, loc)
 
         values = self.visit_and_materialise(expr.value)
 
         return ir.BytesEncode.maybe(
             values=values,
             values_type=value_ir_type,
-            encoding=encoding,
+            encoding=encoded_ir_type.encoding,
```
iglosiggio: You are doing a whole type conversion and then only reading the `encoding` field?

Does this conversion have some useful side-effect that should be documented? (Like populating a cache)

iglosiggio (approved):
Changes look ok. Should we have a similar change on `puya-ts`?

src/puya/ir/builder/main.py
boblat: It has been changed to use `wtype_to_encoding` function instead.

PR #652, opened 2026-02-09
Feat/support String in assert msg

Argimirodelpozo:
This PR allows for support of `algopy.String()` strings as error messages in `assert` statements.
The error messages provided in this way will be logged in a fashion similar to:
```
//(...)
// assert condition evaluation result is on top of stack
bnz cont
pushbytes "error message" //or a sequence of opcodes equivalent to pushing the message to stack (see below)
log
err
cont:
// rest of teal code
//(...)
```
Warnings will be emitted to let developers know they are doing this, the associated cost and confirming this is their intent.
Furthermore, the requirement for the error messages to be compile time constants is relaxed.
If the `algopy.String()` is a compile time constant, it will also be used as a comment to allow for arc56 support.
If not, another warning will be emitted, it will be built and will just be logged as exemplified above.

The motivation behind this change is allowing for compiler support for ARC65 error strings.
A following PR will provide a function to build ARC65 compliant strings.

### DIFF
--- a/test_cases/logged_errors/contract.py
+++ b/test_cases/logged_errors/contract.py
@@ -1,0 +1,14 @@
+from algopy import String, UInt64, arc4, op
+from algopy.arc4 import ARC4Contract
+
+
+class ArrayAccessContract(ARC4Contract):
+    @arc4.abimethod
+    def test_branching_array_call(self, arg: UInt64) -> None:
+        assert arg != 0, "regular arc56 error not logged"
+        assert arg != 1, String("this error should be logged with warnings")
+
+        non_const_string: String = String(str(op.itob(arg))) + String(
+            "non constant string should be logged with warnings"
+        )
+        assert arg != 2, non_const_string


Argimirodelpozo:
> Am I correct in saying that after this update all existing asserts will also result in pushbytes; log? If so we should probably think about a way to disable that behavior due to the extra bytes it would add.

It should not, as currently asserts are of the form:
```
assert condition, str("a compile time constant string")
```
whereas something like
```
assert condition, algopy.String("a compile time constant string")
```
which is an utf-8 encoded, non-arc4 string, would error.

So the "hole" in the implementation of assert is already there. This would just enable those to work, while making their behavior a logged error as default (otherwise you can use `str()` and keep the usual behavior).

If it also happens to be a compile time constant, then we extract the value and _also_ use it in regular arc56 fashion.

If it is not (or its construction is complex enough that we can't figure out it is), we just log it and you loose arc56 compatibility.
Both those features come with warnings, which will be suppresed if you build the `algopy.String()` using an ARC65 helper function, to be added in an upcoming PR.

Argimirodelpozo:
As per discussed with the team, closing in favor of #657.
This implementation idea would not have worked quite as nicely in the puya-ts frontend.
Having non-compile time constants as error messages may be revisited in a future design effort if there is demand by users.

PR #653, opened 2026-02-09
feat: Add client generation for typescript

iglosiggio:
Note: this is a first draft made by copy-pasting the already available code. We still need to polish things up a bit and validate that everything works on `puya-ts`

## Proposed Changes

- Add client generation as typescript code. This facilitates abi calls from `puya-ts` contracts to `puyapy` contracts
- [TO-DO]: Move client generation from `puyapy` to `puya`: This will allow it to be reused by both `puyapy` and `puya-ts`
- [TO-DO]: Add compiler flags for "generate only the python client" and "generate only the typescript client". The currently available flag will generate both
- [TO-DO... later]: Set-up a mechanism for validating both the generated python and typescript clients (for example: check that the generated code typechecks)
- [TO-DO... on `puya-ts`]: Add the new flags to `puya-ts` too

## Additional comments

The current implementation mainly allows us to use abi calls easily. We can eventually devise a mechanism where every contract class has a `Contract.from_app_id` method that instances a contract proxy (this means a dev can write `my_contract.transfer(10)` instead of `arc4.abi_call(MyContract.transfer, 10, app_id=...)`). This looks like a useful way to interact with well-defined contract families like ARC-20 contracts.


iglosiggio:
closing in favor or https://github.com/algorandfoundation/puya-ts/pull/329

PR #657, opened 2026-02-18
Feat/on chain logged errors

Argimirodelpozo:
This PR introduces support for logged assertions (and logged errors) on chain. You may see [ARC65](https://dev.algorand.co/arc-standards/arc-0065/) for more information on the formatting requirements and recommendations of emitted errors.

## Proposed Changes

- added: 
`logged_assert(condition: bool, code: str, message: str | None, prefix: typing.Literal["AER", "ERR"] = "ERR")` 
and similarly 
`logged_err(code: str, message: str | None, prefix: typing.Literal["AER", "ERR"] = "ERR")`. 

Note error messages and codes may not contain ":" as this is used as domain separator when building the ARC65 error string. Also, since building said string happens at compile time, this is compatible with currently existing clients.

- changes to the backend to support both functions. The compilation of said functions now generates essentially
```
//[if logged_assert(), condition branching goes here]
pushbytes "ERR:code:message"
log
err
```

- validations to accomodate to the standard MUSTs and SHOULDs, plus tests to support this

### DIFF
--- a/src/puyapy/awst_build/eb/arc65_error.py
+++ b/src/puyapy/awst_build/eb/arc65_error.py
@@ -105,7 +105,6 @@ def _resolve_error_message(
 
     # code strict properties validation
     if code is None:
-        # should never get here because of type checking and expect.simple_string_literal(.)
         logger.error("error code is mandatory in logged errors", location=location)
         return None
     elif ":" in code:
@@ -118,7 +117,7 @@ def _resolve_error_message(
         return None
 
     # prefix strict properties validation
-    # (note: prefix should already be validated by mypy typing check)
+    # (note: prefix should already be validated by mypy type checking)
     if prefix not in _VALID_PREFIXES:
         logger.error(
             "error prefix must be one of AER, ERR",
--- a/tests/test_expected_output/logged_errors.test
+++ b/tests/test_expected_output/logged_errors.test
@@ -180,3 +180,32 @@ class MyContract(Contract):
 
     def clear_state_program(self) -> bool:
         return True
+
+## case: test_logged_err_missing_code
+
+from algopy import Contract, Txn, logged_err
+
+
+class MyContract(Contract):
+    def approval_program(self) -> bool:
+        if Txn.num_app_args > 0:
+            logged_err() # type: ignore[call-arg] ## E: missing required positional argument(s): error_code \
+                                                   ## E: error code is mandatory in logged errors
+        return True
+
+    def clear_state_program(self) -> bool:
+        return True
+
+## case: test_logged_assert_missing_code
+
+from algopy import Contract, Txn, logged_assert
+
+
+class MyContract(Contract):
+    def approval_program(self) -> bool:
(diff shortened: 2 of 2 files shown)


src/puyapy/awst_build/eb/arc65_error.py
```diff
@@ -0,0 +1,143 @@
...
+        else None
+    )
+    prefix = (
+        expect.simple_string_literal(arg_mapping["prefix"], default=expect.default_none)
+        if "prefix" in arg_mapping
+        else "ERR"
+    )
+
+    # code validation
+    if code is None:
```
daniel-makerx: What do you think about doing the ARC-65 validation either on the node or in the LoggedErrorsValidator so this can be reused across front ends?

daniel-makerx: Doing the warning checks in the AWST validator sounds good to me.

If we are going with `error_message: str | None` then doing the error checks here makes sense.

If we did introduce a more structured type for ARC-65 e.g. `error_message: str | ARC65Error | None` then it would make sense for any validation of the ARC65Error to go on that.

cusma (commented):
Overall this looks great!

I was wondering if the complier should be a bit more opinionated on the error "message". Today the program size is the main bottleneck for AVM devs, so I was wondering if Puya should facilitate "good manners" when it comes to error messages length or format.

ARC-65 _recommends_ that the total error length (`ERR:<code>:<message>`) should not be exactly equal to `8` or `32` bytes (as this would be encoded by the Algod API as `uint64` or `address`). ARC-65 _recommends_ also `camelCaseing`, but it would be too much for the compiler, maybe just a cap on the total length? I have mixed feelings.

src/puyapy/awst_build/eb/arc65_error.py
```diff
@@ -0,0 +1,143 @@
...
+        logger.error("error code is mandatory in logged errors", location=location)
```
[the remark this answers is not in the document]
Argimirodelpozo: Actually for this one I feel its kind of difficult to trigger as:
- typechecking would catch if you don't have a code and
- if said type is not an `str`, the `expect.simple_string_literal(.)` above does catch it. 

Still I left a comment above it stating how this should not happen.

Argimirodelpozo:
> I was wondering if the complier should be a bit more opinionated on the error "message". Today the program size is the main bottleneck for AVM devs, so I was wondering if Puya should facilitate "good manners" when it comes to error messages length or format.


Added all ARC-65 recommendations as warnings! This includes camelCase as well as 8/32 bytes (of the final message string) and also size (for which I arbitrarily decided 64 bytes was a _long_ enough message). This is also pointed out in the docstring for the `logged_assert()` and `logged_err()` functions. I guess capping / forbidding seemed too drastic to me. We can always promote them into errors later, or remove them altogether, according to people's adoption and usage of the feature.

daniel-makerx (approved):

PR #658, opened 2026-02-23
refactor: move type mapping logic for emit function to ir layer

boblat:


src/puya/ir/builder/main.py
```diff
@@ -1441,100 +1431,112 @@ def visit_emit(self, expr: awst_nodes.Emit) -> TExpression:
...
-            maybe_args = ()
-
         if maybe_returns is not None:
             logger.error(
                 "event signatures cannot include return types",
                 location=expr_loc,
             )
 
-        if maybe_args is not None and len(expr.values) != len(maybe_args):
+        if maybe_args_str:
```
daniel-makerx: Combine these blocks? E.g.

```python
        if maybe_args_str:
            maybe_args = tuple(types.split_tuple_types(maybe_args_str))
            if len(maybe_args) != len(expr.values):
                logger.error(
                    f"expected {len(maybe_args)} ABI arguments",
                    location=expr_loc,
                )
                return None
            fields = [types.abi_name_to_wtype(arg) for arg in maybe_args]
        else:
            fields = [arg.wtype for arg in expr.values]
```

PR #660, opened 2026-02-25
Support arguments in logic signatures

Argimirodelpozo:
This PR brings support for arguments in logic signatures (similar to how we handle arguments in regular subroutines).
It also adds the ability to opt into and out of validation through a decorator field in `@logicsig`.

## Proposed Changes
- remove guards preventing arguments in logicsigs at the AWST level
- before program building, construct a "frame" for the arguments (in IR, but still in AWST form). This frame unpacks all arguments from `op.arg(N)` values, decoding them to the appropriate `WType`, and validating if the arg. types are validatable.
- tests attempting thorough evaluation of different kinds of arguments.
- tests showing encoding validation works as expected.


tests/test_logic_sig.py
```diff
@@ -144,3 +151,135 @@ def test_pre_approved_sale(
...
+# Invalid OverwriteStruct, size of 3 bytes.
+# This will corrupt the bool field if it goes through (make it false).
+# Should not pass encoding validation, and should fly under the radar in `unsafe_disabled`.
+
+
+def _build_complex_args(*, overwrite_struct: bytes = _VALID_OVERWRITE_STRUCT) -> list[bytes]:
+    simple_struct = (1).to_bytes(8, "big") + (2).to_bytes(8, "big")
+    nested_struct = b"\x03" + simple_struct
+    return [
+        (42).to_bytes(8, "big"),  # arg0:  UInt64
```
daniel-makerx: We can use algokit utils to do the ARC-4 encoding, a few other onchain tests have a utility function `_arc4_encode`, might be worth moving that into `tests.utils` and using it here

daniel-makerx (approved):
Looks great!

tests/test_logic_sig.py
daniel-makerx: 🥷

PR #691, opened 2026-04-13
Fix miscomputation of identity division when operand could be zero

### LINKED ISSUE #690: `x // x` identity fold silently eliminates division-by-zero panic
The intrinsic simplifier unconditionally folds `x // x` to `1` for both uint64 and biguint when both operands are the same SSA register. This is only valid when `x != 0` — the AVM panics on division by zero, so the fold silently eliminates a runtime error.

### Reproduction

```python
from algopy import Contract, UInt64, subroutine

class Example(Contract):
        assert self_div(UInt64(5)) == 1
        assert self_div(UInt64(0)) == 1  # expected: panic; actual: returns 1
        return True

        return True


@subroutine(inline=False)
def self_div(x: UInt64) -> UInt64:
    return x // x
```

Calling a miscompiled `self_div(0)` returns `1` instead of halting, which could cause contracts to proceed past checkpoints that should have panicked.

Destructured IR at `-O1` shows the fold:
```
subroutine self_div(x: uint64) -> uint64:
    block@0:
        return 1u
```

Executing at `-O1` / `-O2` succeeds and returns `1` (wrong). At `-O0`, the division is preserved and the AVM panics with `/ 0`.

The same reproduction applies to `BigUInt` by simple replacement in the above contract.

achidlow:
Fixes #690


daniel-makerx (approved):

src/puya/ir/optimize/intrinsic_simplification.py
```diff
@@ -1510,12 +1509,11 @@ def _try_simplify_bytes_binary_op(
                 c = 1
             case AVMOp.neq_bytes | AVMOp.neq | AVMOp.lt_bytes | AVMOp.gt_bytes:
                 c = 0
-            case AVMOp.div_floor_bytes:
-                c = 1
             case AVMOp.bitwise_xor_bytes:
                 c = 0
             case AVMOp.bitwise_and_bytes | AVMOp.bitwise_or_bytes:
                 c = a
+            # AVMOp.div_floor_bytes: "\x01", but need to guard against division by zero
```
iglosiggio, line 1520: Should we add another issue for re-introducing these optimizations with the correct guards?

iglosiggio (approved):

src/puya/ir/optimize/intrinsic_simplification.py
achidlow: Maybe? In theory the GVN pass I'm working on could tell you if the value is guaranteed to be non-zero with a bit of a work (through tracking dominating asserts of the VN but also dominating asserts of the VN > 0 or 0 < VN). Sparse conditional constant propagation could be used to track guarantees of non-zero-ness due to branching potentially. 

But as with everything, there is a trade off in terms of performance but also maintenance - and I'm not sure the trade-offs are in favour here just for optimising away `(a // a)`.

However, if the work was being done in the context of a larger suite of optimisations that rely on tracking non-zero-ness, then yeah. But I wouldn't put this on the backlog just by itself - feel free to disagree of course.

(Edit: I'm going to resolve this so that the bug fix can be merged, but let's continue this conversation.)

PR #693, opened 2026-04-13
Fix biguint XOR identity fold result byte-length

### LINKED ISSUE #692: biguint `x ^ x` identity fold loses byte-width of result
The intrinsic simplifier folds biguint `x ^ x` (AVM `b^`) to the integer constant `0`, which encodes as empty bytes. But the AVM `b^` op produces a zero-filled byte array **of the same length as its inputs**, not an empty array. The fold changes observable byte-width, breaking any downstream code that depends on the result length.

### Reproduction

```python
from algopy import BigUInt, Contract, op, subroutine

class Example(Contract):
        x = BigUInt(2**32 - 1)  # 4 bytes: 0xffffffff
        result = self_xor(x)
        assert result.bytes == op.bzero(x.bytes.length)  # expected: 4 zero bytes
        assert result.bytes.length == x.bytes.length      # expected: 4, actual (miscompiled): 0
        return True

        return True

@subroutine(inline=False)
def self_xor(x: BigUInt) -> BigUInt:
    return x ^ x
```

Destructured IR at `-O1` shows the fold:
```
subroutine self_xor(x: biguint) -> biguint:
    block@0:
        return 0b
```

Executing at `-O1` / `-O2` fails the length assertion. At `-O0`, the xor is preserved and the contract succeeds.

achidlow:
Fixes #692


src/puya/ir/optimize/intrinsic_simplification.py
```diff
@@ -1512,10 +1512,10 @@ def _try_simplify_bytes_binary_op(
                 c = 0
             case AVMOp.div_floor_bytes:
                 c = 1
-            case AVMOp.bitwise_xor_bytes:
-                c = 0
             case AVMOp.bitwise_and_bytes | AVMOp.bitwise_or_bytes:
                 c = a
+            # AVMOp.bitwise_xor_bytes: zero-filled array of same length, so can
```
iglosiggio, line 1519: Same comment as in #691. We should log re-adding this properly (if possible)

iglosiggio (approved):

src/puya/ir/optimize/intrinsic_simplification.py
achidlow: Similar to [my comment there](https://github.com/algorandfoundation/puya/pull/691#discussion_r3076652839), I'm not sure if it's justifiable just for this one case of optimising `a ^ a` where `a` is biguint. However, if it was being done as part of a larger piece of work about tracking known lengths, I think that could actually be quite valuable - especially given the new (relatively speaking) automatic length assertions!

(Edit: I'm going to resolve this so that the bug fix can be merged, but let's continue this conversation.)

daniel-makerx (approved):

PR #694, opened 2026-04-16
Fix: bug in getbit intrinsic simplification folding (with uint64 and bytes)

Argimirodelpozo:
This PR fixes two small bugs in intrinsic simplification:

- When attempting to fold a `getbit` out of bounds index access on a uint64, the compiler would silently fold it to a constant uint64 0.

- In the same scenario but on a bytes constant, it'd cause a compiler crash.

Now both scenarios compile, however they produce a piece of code that will panic on runtime.
It might be sensible to introduce warnings for these scenarios, however this is out of scope for this fix.

### DIFF
--- a/changelog.d/20260415_205953_argimirodelpozo_getbit_bytes_fold_oob.md
+++ b/changelog.d/20260415_205953_argimirodelpozo_getbit_bytes_fold_oob.md
@@ -1,0 +1,45 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category.
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+
+### Fixed
+
+- Fixed a miscompilation where `getbit` accessing an out of bounds index in a constant `UInt64` would be folded into a constant 0 integer. The AVM should panic in this scenario.
+
+- Fixed a compiler crash when attempting to constant-fold a `getbit` out of bounds index access into a bytes constant.
+
+
+<!--
+### Security
+
+- A bullet item for the Security category.
+
+-->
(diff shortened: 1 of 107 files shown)


tests/onchain/test_regression_tests.py
```diff
@@ -27,3 +27,16 @@ def test_div_identity_fold_biguint(deployer_o: Deployer) -> None:
 
 def test_xor_identity_fold_biguint(deployer_o: Deployer) -> None:
     deployer_o.create_bare(TEST_CASES_DIR / "regression_tests" / "xor_identity_fold_biguint.py")
+
+
+def test_gebit_bytes_fold_oob(deployer_o: Deployer) -> None:
```
achidlow:
```suggestion
def test_getbit_bytes_fold_oob(deployer_o: Deployer) -> None:
```

changelog.d/20260415_205953_argimirodelpozo_getbit_bytes_fold_oob.md
```diff
@@ -0,0 +1,45 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+
+### Fixed
+
+- Fixed a miscompilation where `getbit` accessing an out of bounds index in a constant `UInt64` would be folded into a constant 0 integer. The AVM should panic in this scenario.
```
achidlow:
```suggestion
- Fixed a miscompilation where `getbit` accessing an out of bounds index in a constant `uint64` would be folded into a constant 0 integer. The AVM should panic in this scenario.
```

test_cases/regression_tests/getbit_fold_oob_uint64.py
```diff
@@ -0,0 +1,11 @@
+from algopy import Contract, UInt64, op
+
+
+class GetBitFoldOOBUInt64(Contract):
+    def approval_program(self) -> bool:
+        # out of bounds bit index access (should be indices 0...63)
+        assert op.getbit(UInt64(0), 64) == UInt64(0)
```
achidlow:
```suggestion
        assert op.getbit(0, 64) == 0
```

test_cases/regression_tests/getbit_fold_oob_bytes.py
```diff
@@ -0,0 +1,11 @@
+from algopy import Bytes, Contract, UInt64, op
+
+
+class GetBitFoldOOBBytes(Contract):
+    def approval_program(self) -> bool:
+        # out of bounds bit index access (8 bits, indices 0...7)
+        assert op.getbit(Bytes(b"\xff"), 8) == UInt64(0)
```
achidlow:
```suggestion
        assert op.getbit(b"\xff", 8) == 0
```

boblat (approved):

Argimirodelpozo:
Removing the explicit types in `getbit` causes an `unsupported comparison between types` error. Will investigate and open a separate PR/issue; preliminarily it seems to be related to bool `compare()` lacking upcasting when dealing with integers.

PR #695, opened 2026-04-16
fix: detect recursive type aliases instead of crashing with maximum recursion depth

boblat:
- Self-referential type aliases (e.g. `Arr = arc4.DynamicArray["Arr"]`) now emit a clear `CodeError` instead of crashing with `RecursionError`
- Added cycle detection via a `_resolving_aliases` frozenset threaded through `type_to_pytype` and `_maybe_parameterise_pytype`

### DIFF
--- a/src/puyapy/awst_build/context.py
+++ b/src/puyapy/awst_build/context.py
@@ -206,6 +206,7 @@ def type_to_pytype(
     source_location: SourceLocation,
     in_type_args: bool = False,
     in_func_sig: bool = False,
+    _resolving_aliases: frozenset[str] = frozenset(),
 ) -> pytypes.PyType:
     loc = source_location
     proper_type_or_alias: mypy.types.ProperType | mypy.types.TypeAliasType
@@ -219,6 +220,7 @@ def type_to_pytype(
         source_location=loc,
         in_type_args=in_type_args,
         in_func_sig=in_func_sig,
+        _resolving_aliases=_resolving_aliases,
     )
     match proper_type_or_alias:
         case mypy.types.TypeAliasType(alias=alias, args=args):
@@ -226,8 +228,15 @@ def type_to_pytype(
                 raise InternalError("mypy type alias type missing alias reference", loc)
             result = registry.get(alias.fullname)
             if result is None:
-                return recurse(mypy.types.get_proper_type(proper_type_or_alias))
-            return _maybe_parameterise_pytype(registry, result, args, loc)
+                if alias.fullname in _resolving_aliases:
+                    raise CodeError("recursive type aliases are not supported", loc)
+                return recurse(
+                    mypy.types.get_proper_type(proper_type_or_alias),
+                    _resolving_aliases=_resolving_aliases | {alias.fullname},
+                )
+            return _maybe_parameterise_pytype(
+                registry, result, args, loc, _resolving_aliases=_resolving_aliases
+            )
         # this is how variadic tuples are represented in mypy types...
         case mypy.types.Instance(type=mypy.nodes.TypeInfo(fullname="builtins.tuple"), args=args):
             try:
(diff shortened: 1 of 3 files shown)


achidlow (approved):

achidlow:
Alright, I tried to break this, but I couldn't:

I tried the following test cases:
```
## case: recursive_type_alias
import typing
from algopy import arc4

# auto-recursion through union
JSON: typing.TypeAlias = int | float | str | bool | None | list["JSON"] | dict[str, "JSON"] ## E: unsupported type-alias format

# auto-recursion through generic param
Arr = arc4.DynamicArray["Arr"] ## E: recursive type aliases are not supported

# mutually recursive aliases
Leaf: typing.TypeAlias = int
Branch: typing.TypeAlias = tuple["Tree", "Tree"] ## E: recursive type aliases are not supported
Tree: typing.TypeAlias = Leaf | Branch ## E: unsupported type-alias format

# mutual recursion through generic params
Arr1 = arc4.DynamicArray["Arr2"] ## E: recursive type aliases are not supported
Arr2 = arc4.DynamicArray["Arr1"] ## E: recursive type aliases are not supported

# recursive generic type
T = typing.TypeVar("T") ## E: unsupported type-form \
                        ## E: unsupported statement type at module level

class LinkedList(arc4.ARC4Contract, typing.Generic[T]):
    head: T
    tail: "LinkedList[T] | None" ## E: type unions are unsupported at this location

# self-referntial classes
class TreeNode(arc4.Struct):
    value: arc4.UInt64
    left: "TreeNode | None" ## E: Unknown type: recursive_type_alias.TreeNode
    right: "TreeNode | None"

# "phantom" recursion
Nested: typing.TypeAlias = T | list["Nested[T]"] ## E: unsupported type-alias format

## case: invalid_recursive_type
from typing import TypeAlias

Bad: TypeAlias = "Bad" ## E: Cannot resolve name "Bad" (possible cyclic definition)  [misc] \
                       ## E: unsupported type-alias format
A: TypeAlias = "B" ## E: Cannot resolve name "B" (possible cyclic definition)  [misc] \
                   ## E: unsupported type-alias format \
                   ## E: Cannot resolve name "A" (possible cyclic definition)  [misc]
B: TypeAlias = "A" ## E: Cannot resolve name "B" (possible cyclic definition)  [misc] \
                   ## E: unsupported type-alias format
```

Notably, we don't support type aliases in user code, so that fails at the statement level. Adding support for `OpExpr` in type aliases there only results in some changes of error messages above from "unsupported type-form" to the recursive type alias message.

LGTM!

tests/test_expected_output/module.test
```diff
@@ -90,6 +90,58 @@ IntAlias: typing.TypeAlias = UInt64
...
+
+# recursive generic type
+T = typing.TypeVar("T") ## E: unsupported type-form \
+                        ## E: unsupported statement type at module level
+
+class LinkedList(arc4.ARC4Contract, typing.Generic[T]):
+    head: T
+    tail: "LinkedList[T] | None" ## E: type unions are unsupported at this location
+
+# self-referntial classes
```
Argimirodelpozo:
```suggestion
# self-referential classes
```

src/puyapy/awst_build/context.py
```diff
@@ -206,28 +206,34 @@ def type_to_pytype(
...
+        if alias_fullname in _resolving_aliases:
+            raise CodeError("recursive type aliases are not supported", loc)
+        result = registry.get(alias_fullname)
+        if result is not None:
+            return _maybe_parameterise_pytype(
+                registry, result, mypy_type.args, loc, _resolving_aliases=_resolving_aliases
+            )
+        # fall through to resolve the proper type, which will do a single expansion of
+        # the type alias, but track that we're now resolving this alias name
+        _resolving_aliases |= {alias_fullname}
```
Argimirodelpozo, line 225: Should this line adding the `alias_fullname` to the set be moved to before the `if result is not None:` line so the call to `_maybe_parameterise_pytype` sees the `alias_fullname` in `_resolving_aliases`? I couldn't come up with an example that breaks its current state, however it'd be correct right? That way it is _deffinitely_ impossible to recurse through the arg. of a generic.

Argimirodelpozo (approved):
Looks good to me! just a nit typo and the question about adding the alias name prior, tho it might be inconsequential 🙃

tests/test_expected_output/module.test
boblat: :+1: Fixed.

src/puyapy/awst_build/context.py
boblat: We did try moving it to before the `if result is not None:`. Doing so would throw `recursive type aliases are not supported` error for [Nested_Tuple = arc4.Tuple\[arc4.Tuple\[arc4.UInt64, arc4.UInt64\], arc4.Tuple\[arc4.String, arc4.String\]\]\]](https://github.com/algorandfoundation/puya/pull/695/changes#diff-e509a850d3ca047e4015de81dce45f7a12a6ad0b06371d1289b30e4cbacaab8bR99).

achidlow: Yeah it needs to be after, because if the base type is in the registry, it's either a standard type we ship (which cannot be recursive), or a type that has fully resolved previously without recursion. However, as per the snippet Bobby posted above, if we add it to resolving types when doing parameterisation, that breaks when parameterisation has nesting of the same type (but not recursion).

PR #698, opened 2026-04-17
Fix OOB setbit uint64 constant fold

### LINKED ISSUE #697: `setbit` on `uint64` constant with a constant index >= 64 gets constant folded
Problems:
- `op.setbit_uint64(0, 64, True)` will be constant-folded to zero, instead of producing a panic.
- `op.setbit_uint64(42, 64, False)` will produce an oversized value, producing a critical compilation failure.

The intrinsic simplifier constant-folds `setbit` on `uint64` values even when the bit index is >= 64 (valid range is 0–63). 

### Full example

```python
from algopy import *

class SetbitContract(ARC4Contract):
    @public
    def crash(self) -> UInt64:
        return op.setbit_uint64(0, 64, True)

    @public
    def wrong(self) -> None:
        assert op.setbit_uint64(42, 64, False) == 42
```

As is, will fail with `critical: Invalid UInt64 value.`, commenting out the `crash` method will reveal the `wrong` method body is optimised away entirely.

achidlow:
Fixes #697

### DIFF
--- a/src/puya/ir/optimize/intrinsic_simplification.py
+++ b/src/puya/ir/optimize/intrinsic_simplification.py
@@ -612,6 +612,8 @@ def _try_fold_intrinsic(
                 models.UInt64Constant(value=index),
                 models.UInt64Constant(value=value),
             ]:
+                if index >= 64:
+                    return None
                 if value:
                     setbit_result = source | (1 << index)
                 else:
--- a/tests/onchain/test_regression_tests.py
+++ b/tests/onchain/test_regression_tests.py
@@ -40,3 +40,8 @@ def test_getbit_uint64_fold_oob(deployer_o: Deployer) -> None:
     # Without the fix, the optimizer folds this to UInt64(0) and the panic is lost.
     with pytest.raises(au.LogicError, match="had error 'getbit index > 63 with Uint'"):
         deployer_o.create_bare(TEST_CASES_DIR / "regression_tests" / "getbit_fold_oob_uint64.py")
+
+
+def test_setbit_uint64_fold_oob(deployer_o: Deployer) -> None:
+    with pytest.raises(au.LogicError, match="setbit index > 63 with Uint"):
+        deployer_o.create_bare(TEST_CASES_DIR / "regression_tests" / "setbit_fold_oob_uint64.py")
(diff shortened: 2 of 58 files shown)


tests/onchain/test_regression_tests.py
```diff
@@ -27,3 +27,12 @@ def test_div_identity_fold_biguint(deployer_o: Deployer) -> None:
...
     deployer_o.create_bare(TEST_CASES_DIR / "regression_tests" / "xor_identity_fold_biguint.py")
+
+
+def test_setbit_uint64_fold_oob(deployer_o: Deployer) -> None:
+    # OOB setbit on uint64 should panic on the AVM.
+    # Without the fix, the optimizer folds setbit(0, 64, 1) to an oversized constant
+    # and setbit(42, 64, 0) silently to 42.
+    with pytest.raises(au.LogicError, match="setbit index") as ex_info:
+        deployer_o.create_bare(TEST_CASES_DIR / "regression_tests" / "setbit_fold_oob_uint64.py")
+    assert ex_info.value.message == "setbit index > 63 with Uint"
```
boblat: should we assert the error message as part of `with(..., match="...")` to match the other tests and remove the extra assertion?

boblat (approved):

tests/onchain/test_regression_tests.py
achidlow: Yep, fixed

src/puya/ir/optimize/intrinsic_simplification.py
```diff
@@ -607,6 +607,8 @@ def _try_fold_intrinsic(
                 models.UInt64Constant(value=index),
                 models.UInt64Constant(value=value),
             ]:
+                if index >= 64:
+                    return None
```
iglosiggio, line 611: Should this produce a compiler warning?

iglosiggio (approved):

iglosiggio (approved):

src/puya/ir/optimize/intrinsic_simplification.py
achidlow: The pattern thus far has been to just not optimise invalid operations, in case they're going to be eliminated. I think this is still appropriate, perhaps a separate enhancement could be to make lower levels emit warnings (e.g. during MIR build), would need to look at doing it consistently across all ops I think.

PR #700, opened 2026-04-17
Fix OOB shl/shr constant fold

### LINKED ISSUE #699: `shl`/`shr` on `uint64` constants with a shift amount >= 64 gets constant folded
The intrinsic simplifier constant-folds `shl` and `shr` even when the shift amount is >= 64, whereas the AVM panics with "shl arg too big" / "shr arg too big".

### Full example

```python
from algopy import *

class ShiftContract(ARC4Contract):
    @arc4.abimethod
    def wrong(self) -> None:
        assert op.shl(1, 64) == 0
        assert op.shr(1, 64) == 0
```

Both asserts are optimised away entirely — the method body becomes a no-op.

achidlow:
Fixes #699

### DIFF
--- a/src/puya/ir/optimize/intrinsic_simplification.py
+++ b/src/puya/ir/optimize/intrinsic_simplification.py
@@ -1464,9 +1464,9 @@ def _try_simplify_uint64_binary_op(
                     c = int(a_const and b_const)
                 case AVMOp.or_:
                     c = int(a_const or b_const)
-                case AVMOp.shl:
+                case AVMOp.shl if b_const < 64:
                     c = (a_const << b_const) % (2**64)
-                case AVMOp.shr:
+                case AVMOp.shr if b_const < 64:
                     c = a_const >> b_const
                 case AVMOp.exp:
                     if a_const == 0 and b_const == 0:
--- a/tests/onchain/test_regression_tests.py
+++ b/tests/onchain/test_regression_tests.py
@@ -45,3 +45,17 @@ def test_getbit_uint64_fold_oob(deployer_o: Deployer) -> None:
 def test_setbit_uint64_fold_oob(deployer_o: Deployer) -> None:
     with pytest.raises(au.LogicError, match="setbit index > 63 with Uint"):
         deployer_o.create_bare(TEST_CASES_DIR / "regression_tests" / "setbit_fold_oob_uint64.py")
+
+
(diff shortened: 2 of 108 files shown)


tests/onchain/test_regression_tests.py
```diff
@@ -27,3 +27,19 @@ def test_div_identity_fold_biguint(deployer_o: Deployer) -> None:
...
 def test_xor_identity_fold_biguint(deployer_o: Deployer) -> None:
     deployer_o.create_bare(TEST_CASES_DIR / "regression_tests" / "xor_identity_fold_biguint.py")
+
+
+def test_shl_fold_oob(deployer_o: Deployer) -> None:
+    with pytest.raises(au.LogicError, match="shl arg too big") as ex_info:
+        deployer_o.create_bare(
+            (TEST_CASES_DIR / "regression_tests" / "shift_fold_oob.py", "ShlFoldOOB")
+        )
+    assert ex_info.value.message == "shl arg too big, (64)"
```
boblat: should we assert the error message as part of `with(..., match="...")` to match the other tests and remove the extra assertion?

boblat (approved):

tests/onchain/test_regression_tests.py
achidlow: Yep, fixed

src/puya/ir/optimize/intrinsic_simplification.py
```diff
@@ -1457,9 +1457,9 @@ def _try_simplify_uint64_binary_op(
                     c = int(a_const and b_const)
                 case AVMOp.or_:
                     c = int(a_const or b_const)
-                case AVMOp.shl:
+                case AVMOp.shl if b_const < 64:
                     c = (a_const << b_const) % (2**64)
-                case AVMOp.shr:
+                case AVMOp.shr if b_const < 64:
```
iglosiggio, line 1462: Again: it feels like this warrants a compiler warning at minimum 😛

iglosiggio (approved):

src/puya/ir/optimize/intrinsic_simplification.py
achidlow: https://github.com/algorandfoundation/puya/pull/698#discussion_r3107897125

PR #702, opened 2026-04-17
Fix OOB replace constant fold

### LINKED ISSUE #701: Constant-folding a `replace` op with all-constant data succeeds when it shouldn't
The intrinsic simplifier constant-folds `replace2` and `replace3` even when the replacement bytes extend past the end of the source. The AVM panics with "replacement end N beyond original length: M", but instead the optimiser grows the array.

### Full example

```python
from algopy import *

class ReplaceContract(ARC4Contract):
    @public
    def wrong(self) -> None:
        assert op.replace(b"", 0, b"abc") == b"abc"
```

The `assert` is optimised away — the method body becomes a no-op. It should either fail to compile or preserve the (guaranteed to fail) `replace` op.

achidlow:
Fixes #701

### DIFF
--- a/changelog.d/20260417_153801_achidlow_replace_fold_oob.md
+++ b/changelog.d/20260417_153801_achidlow_replace_fold_oob.md
@@ -1,0 +1,41 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category.
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+### Fixed
+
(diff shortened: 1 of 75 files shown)


boblat (approved):

iglosiggio (approved):
Same comment as with all the other ones. For completime-known panics I would like to have a warning/error. If we want to still compile the code then I would like to optimize the code away into an `err` (so dead code elim can trim out whatever comes after it).

achidlow:
> Same comment as with all the other ones. For completime-known panics I would like to have a warning/error. If we want to still compile the code then I would like to optimize the code away into an err (so dead code elim can trim out whatever comes after it).

This PR and all the others follow the established pattern in intrinsic simplification pass, which is to not modify potentially invalid operations. This is because it might have become invalid due to another optimisation pass, but before it is removed by dead code elimination. Whether this is a legitimate concern or only a hypothetical one is up for discussion.

We could also add warnings *after* all IR optimisations have been done, e.g. when building MIR. Or even later, at TEAL assembly point.

Inserting an `err` op is not the way to go I think - it means splitting blocks here, making this pass even more complicated. 

Some other sort of [poison value](https://llvm.org/devmtg/2020-09/slides/Lee-UndefPoison.pdf) might work, but I'm not sure this addition would pay off for us.

PR #707, opened 2026-04-17
Fix && and || constant fold for non-boolean operands

achidlow:
The constant folding for AVM `&&` and `||` used Python's `int(a and b)` / `int(a or b)`, which returns the operand value rather than 0 or 1 when operands are non-boolean. The AVM ops always return 0 or 1.

Neither Algorand Python nor puya-ts currently emit non-boolean operands for these ops, so this is future-proofing rather than a fix for an active bug — no issue or changelog entry.


src/puya/ir/optimize/intrinsic_simplification.py
achidlow, line 1410: TODO: consider if these need changing as well. Might need to only consider `||` when in a `bool_context`.

achidlow: Yes it does need changing - AWST test case added for negative case. Python test cases add for positive case / maintaining coverage.

daniel-makerx (approved):

boblat (approved):

PR #711, opened 2026-04-20
fix: reject intrinsic op variants with immediate when literal argument value exceeds its range

### LINKED ISSUE #709: Variants of op codes with unit8 immediates are incorrectly used even when the parameter exceeds uint8 bound
Mapping code responsible for mapping algopy op to AVM op code (e.g. `algopy.op.replace` -> `replace2` | `replace3`) favours the op code with immediates (e.g. `replace2` over `replace3`) even when the parameter value exceeds the bounds of the immediate. 

The following example incorrectly fails to compile with `Invalid immediate, expected value between 0 and 255` error. 

**Full example**

```python
from algopy import Contract, op


class ReplaceOpSelection(Contract):
        assert op.replace(b"\x01" * 300, 256, b"abc") == b"abc"
        return True

        return True

```

boblat:
Fixes [#709](https://github.com/algorandfoundation/puya/issues/709).

- when building IR from AWST, always pick the stack-arg variant and let the IR optimiser promote stack args to immediates when the values actually fit.

- add checks on literal value <= 255 to immediate promotions for `gitxnas`, `itxnas`, `loads/stores` in IR optimiser 

- add immediate promotion cases in IR optimiser for `gaids`, `gloads`, `gloadss`, `txnas`, `gtxns`, `gtxnas`, `gtxnsa`, `gtxnsas` so the new AWST behavior still produces the compact form when possible.

*Note*:  intrincis simplification optimiser does not run for O0 so O0 stats are up.


Argimirodelpozo (commented):
So I think that for any op. that would unconditionally fail with something out of bounds for a certain type (gaid, gload, etc.) we should not let that compile but instead emit an error directly. The argument for the different handling vs. what happened with intrinsic folding is that with intrinsic folding it might not be obvious for a dev. that the folding is happening (something might end up foldable after some other opts.) while in this case it'd be a result of direct user input.
Also in this case we wouldn't have the -O0 vs -O1/-O2 divergent behavior, so I'd say lets be aggresive in those instances and fail compilation.

daniel-makerx:
> So I think that for any op. that would unconditionally fail with something out of bounds for a certain type (gaid, gload, etc.) we should not let that compile but instead emit an error directly. 

When you say out of bounds here, are you referring to the protocol limit of 16 for number of transactions in a group? If so is that a bound that could ever realistically be changed at the protocol level?

Argimirodelpozo:
>When you say out of bounds here, are you referring to the protocol limit of 16 for number of transactions in a group? If so is that a bound that could ever realistically be changed at the protocol level?

Yes (the `T` param. for `gaid` and `gload`) and also talking about the scratch limit in `gload` (the `I` param bounded by 255).
I'd say both of these are very stable protocol bounds. Not that they _absolutely_ couldn't, but if they were to be changed it'd constitute a major breaking change (and I don't think they have [ever changed](https://github.com/algorand/go-algorand/blame/1ad715d077326f7bde11a896b084351b712e6302/config/consensus.go#L997) since their introduction), so IMO they are safe to assume hard constants (unlike something like txn fields that I'd say is more subject to change from time to time e.g. when `RejectVersion` was introduced recently).

achidlow (approved):

Argimirodelpozo (approved):
Looks good to me!

PR #712, opened 2026-04-20
Feat/bool compare int upcast

Argimirodelpozo:
## Proposed Changes

Implement bool upcasting when comparing with an integer literal


src/puyapy/awst_build/eb/bool.py
```diff
@@ -78,15 +78,19 @@ def bool_eval(self, location: SourceLocation, *, negate: bool = False) -> Instan
...
+        if other.pytype == pytypes.BoolType:
+            cmp_expr = NumericComparisonExpression(
+                source_location=location,
+                lhs=self.resolve(),
+                operator=NumericComparison(op.value),
+                rhs=other.resolve(),
+            )
+            return BoolExpressionBuilder(cmp_expr)
+        if other.pytype == pytypes.IntLiteralType:
+            from puyapy.awst_build.eb.uint64 import _upcast_bool
```
boblat: minor: perhaps, we should remove `_` prefix from `_upcast_bool` function name.

boblat (approved):

src/puyapy/awst_build/eb/bool.py
achidlow: Yeah I would suggest renaming to `upcast_bool_to_uint64` and moving it to `awst_build.eb._utils`

achidlow (approved):

Issue #715, opened 2026-04-27
create method on Box class indicates UInt64 | int as types but fails to compile if using int


atsoc1993:
### Typing for `Box.create(size=x)` method indicates it supports `UInt64` or `int` but compiler errors if using an `int` and not a `UInt64`

```
test.py:16:25 error: unexpected argument type
                            box.create(size=8)
```

### Steps to Reproduce

algokit compile py <file_name.py>

```
from algopy import ARC4Contract, UInt64, Box
from algopy.arc4 import abimethod

class Test(ARC4Contract):
    def __init__(self) -> None:
        self.box_test = Box(UInt64, key=b'asd')

    @abimethod
    def test_box(self) -> None:
        box = self.box_test
        box.create(size=8)
        box.value = UInt64(35)
```

### Expected Behavior

Compile Successfully even if using an int

### Screenshots

<img width="380" height="96" alt="Image" src="https://github.com/user-attachments/assets/c8709b45-f215-4892-90ba-1d5c8b54ea16" />

### Device Info

Windows 11

### Additional Context

info: using puyapy version 5.8.1

Argimirodelpozo:
Hi Leo, thanks for reporting the issue! I'm working on a fix on PR #716 and should be fairly straightforward.

For some broader context: when dealing with a box typed as `UInt64` (or any statically sized type for that matter), the compiler already knows the size at compile time. Because of that, passing an explicit size during creation doesn’t really make sense and should likely be disallowed. There might be rare migration-related edge cases where it’s useful, but none come to mind right now. We’ll take a closer look at that in a follow up PR.

On the other hand, for dynamically sized types, the size is usually computed at runtime. In those cases, you'd end up using `UInt64` values in the calculation, which is why this issue hadn’t surfaced before.

atsoc1993:
Awesome — no worries, I think it absolutely could be added complexity to `Box` usage and can't think of edge-cases either, 

`BoxMap` should be the go-to, just need to explicitly warn against size constraints vs. `Box` usage. Should someone use `Box` without needing to the learning curve is a bit higher there.

Will leave it in your hands and thank you for the quick response.

Argimirodelpozo:
Closing the issue as it has been merged in #716. 
@atsoc1993 thanks for the report!

PR #716, opened 2026-04-28
fix: allow int literals as size arguments for box creation

Argimirodelpozo:
Fixes issue #715 by allowing literals in box create size args.

@iglosiggio raised the question as to whether we should even support the size argument for statically typed box creation, which I agree we should disallow in a separate PR (still, no harm in supporting literals wherever the `size` argument would make sense).

Note that create in boxRefs goes through a separate code path where the literal resolution was already in place. Still left a test exercising said resolution.

### DIFF
--- a/changelog.d/20260428_142047_argimirodelpozo_box_create_allow_int_literal.md
+++ b/changelog.d/20260428_142047_argimirodelpozo_box_create_allow_int_literal.md
@@ -32,7 +32,7 @@ For top level release notes, leave all the headers commented out.
 
 ### Fixed
 
-- Allowed supporting integer literals for the `size` argument in `Box.create`
+- Allowed integer literals as `size` arguments in `Box.create`
 
 
 <!--
--- a/changelog.d/20260428_142047_argimirodelpozo_box_create_allow_int_literal.md
+++ b/changelog.d/20260428_142047_argimirodelpozo_box_create_allow_int_literal.md
@@ -1,0 +1,43 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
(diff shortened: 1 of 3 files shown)


src/puyapy/awst_build/eb/storage/box.py
```diff
@@ -249,14 +249,19 @@ def call(
...
     ) -> InstanceBuilder:
-        arg = expect.at_most_one_arg_of_type(args, [pytypes.UInt64Type], location)
-        if arg is not None:
-            size = arg.resolve()
-        else:
-            size = SizeOf(
-                size_wtype=self.content_type.checked_wtype(location),
-                source_location=location,
-            )
+        arg = expect.at_most_one_arg_of_type(
```
boblat, line 252: it now inadvertently allows `box.create(size=True)` which previously would have thrown `unexpected argument type` error.

changelog.d/20260428_142047_argimirodelpozo_box_create_allow_int_literal.md
```diff
@@ -0,0 +1,43 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+
+### Fixed
+
+- Allowed supporting integer literals for the `size` argument in `Box.create`
```
boblat: minor: I think we should drop `supporting` from the sentence as it is duplicating `Allowed`.

src/puyapy/awst_build/eb/storage/box.py
Argimirodelpozo: Oh yeah! It was falling through the match. Thanks, added explicit `None` handling and an explicit error on the fallthrough case.

boblat (approved):

PR #717, opened 2026-05-04
Stable `TMPL_` byte sizes

iglosiggio:
## Proposed Changes

This PR modifies the way we generate bytecode so a statically-known amount of bytes is used whenever possible. For uint64 values this makes our constants bigger (unless `-O2` is chosen).

- Add a `num_bytes` property to the TemplateVar definition in both the MIR an TEAL layers
- Add a `_TemplateInt` marker class that allows the encoding layer to distinguish regular constants from template variable constants
- Update the compiled code

## TO-DO

- Check that these changes work properly with cblock offsets when debugging
- Implement the `_TemplateInt` change in a cleaner way

### Outside scope

- Fix typechecking issues when substituting template variables (AK-1224)


src/puya/ussemble/assemble.py
```diff
@@ -326,10 +346,16 @@ def _encode_op(op: models.AVMOp, *, get_label_offset: Callable[[models.Label], i
...
 
 def _encode_varuint(value: int) -> bytes:
+    assert value < (1 << 64), "Value does not fit an uint64!"
+
+    is_template = isinstance(value, _TemplateInt)
     bits = value & 0x7F
     value >>= 7
     result = b""
-    while value:
+    for i in range(0, 9):
```
daniel-makerx, line 355: Should this be 10 bytes? If each byte can carry 7 bits then 7 * 9 is 63 bits, one short of 64

PR #724, opened 2026-05-07
[fix] PuyaPy: `+=` with `ImmutableArray` inside another structure.

achidlow:
Currently, this fails to compile, because a `SingleEvaluation` node is not a valid assignment target.

The intention behind the current code is correct, but it will only ever succeed if the `.single_eval()` call is a no-op.

The problem that needs to be solved though is demonstrated in the added test case - if there are expressions with side-effects nested within the otherwise-valid `Lvalue` target.

The approach here attempts to balance minimal change whilst maintaining the existing defensive validation checks. We could avoid adding the `is_assignment` flag by removing the `wtype.immutable` check in the AWST validation. This would make sense in a way, since the front-end sets this flag and also constructs the node, presumably it's validating that the source language operation is valid.


src/puya/awst/validation/immutable.py
```diff
@@ -33,7 +33,7 @@ def visit_array_pop(self, expr: awst_nodes.ArrayPop) -> None:
 
     def visit_array_extend(self, expr: awst_nodes.ArrayExtend) -> None:
         super().visit_array_extend(expr)
-        if expr.base.wtype.immutable:
+        if expr.base.wtype.immutable and not expr.is_assignment:
```
iglosiggio, line 36: The text of this line feels misleading. Can you add a comment explaining what is meant?

src/puyapy/awst_build/eb/native/immutable_array.py
```diff
@@ -122,17 +123,11 @@ def augmented_assignment(
...
-            value=ArrayConcat(
-                left=lhs,
-                right=rhs.resolve(),
-                source_location=location,
-            ),
-            source_location=location,
+        extend = ArrayExtend(
+            base=self.resolve(), other=rhs.resolve(), is_assignment=True, source_location=location
         )
+        return ExpressionStatement(expr=extend)
```
iglosiggio, line 130: I understand __why__ this works but it _feels_ wrong to me (if we had some kind of reference semantics I think this breaks the idea that the value of `lhs` is immutable).

src/puya/awst/validation/immutable.py
Argimirodelpozo: If I understood the meaning correctly, maybe a slight renaming of the flag could help? `is_aug_assignment`? `aug_assign_immutable_lvalue`? Idk

achidlow: Yeah I agree - I'm not a fan of this approach overall, it's just the least invasive one I could see to fix the bug for now without introducing additional issues/fragility. Happy for you @iglosiggio to take this one on and explore the alternative solution space, as discussed.

PR #725, opened 2026-05-08
`MacroOps` implementation

iglosiggio:
## Proposed Changes

Add an abstraction that allows us to describe common sequences of generated code. Eventually we want to move all our "templated" code there, allowing it to be inline/outlined depending on the optimization priorities.

## To be improved later

- Add additional logic that enables inlined code to have useful source locations. 
- Compile macro ops to something other than a subroutine call. This should allow us to optimize and reason about our code at the macro-level.


test_cases/compile/out_O2/Hello.approval.teal
```diff
@@ -99,3 +85,18 @@ main_update@14:
...
+    frame_dig -1
+    intc_0 // 0
+    extract_uint16 // on error: invalid array length header
+    pushint 2
+    +
+    frame_dig -1
+    len
+    ==
+    assert // invalid number of bytes for arc4.dynamic_array<arc4.uint8>
+    retsub
```
iglosiggio, line 102: The way proto and frame_dig are encoded forces this to be a bit bigger than the original code. If we had a way to compile small functions in a protoless way this should be smaller:
```teal
%Validate__arc4.dynamic_array<arc4.uint8>:
    dup
    intc_0 // 0
    extract_uint16 // on error: invalid array length header
    pushint 2
    +
    swap
    len
    ==
    assert // invalid number of bytes for arc4.dynamic_array<arc4.uint8>
    retsub
```

examples/amm/out_O2/ConstantProductAMM.approval.assembly-report
```diff
@@ -59,19 +59,19 @@
...
+| 153 | 23                                               |     intc_1 // 1                                                  |                                                                                        | INTERNAL:1                                                                                       |
+| 154 | 12                                               |     ==                                                           |                                                                                        | INTERNAL:1                                                                                       |
+| 155 | 44                                               |     assert // invalid number of bytes for arc4.uint8             |                                                                                        | INTERNAL:1                                                                                       |
```
iglosiggio, line 66: Should we _fix_ this right away? When they are outlined them no `SourceLocation` makes sense. But when we inline them it makes sense to inherit the location from the callsite.

PR #733, opened 2026-05-18
Adds new TEAL optimisation that eliminates cover/uncover by reordering ops

daniel-makerx:
Currently this optimization is only at O2, primarily because it is reordering ops. However it could easily be done at O1 as well


src/puya/teal/optimize/move_shuffled_ops.py
```diff
@@ -0,0 +1,295 @@
...
+def _is_barrier(*, op_to_move: models.TealOp, maybe_barrier_op: models.TealOp) -> bool:
+    """Returns true if maybe_barrier_op invalidates the value op_to_move would read."""
+    moving_op_code = op_to_move.op_code
+    if moving_op_code == "frame_dig":
+        return (
+            maybe_barrier_op.op_code == "frame_bury"
+            and maybe_barrier_op.immediates == op_to_move.immediates
+        )
+    if moving_op_code == "load":
+        return maybe_barrier_op.op_code == "stores" or (
```
Argimirodelpozo: I think maybe callsub should be a barrier here too? (e.g. a subroutine that stores to the same slot as the load).

src/puya/teal/optimize/move_shuffled_ops.py
```diff
@@ -0,0 +1,295 @@
...
+            maybe_barrier_op.op_code == "store"
+            and maybe_barrier_op.immediates == op_to_move.immediates
+        )
+    if moving_op_code in _ITXN_RESULT_OPS:
```
Argimirodelpozo, line 269: And here as well I think? otherwise a callsub could issue an inner and overwrite results of itxn result ops.

src/puya/teal/optimize/main.py
```diff
@@ -119,6 +120,9 @@ def _optimize_block(block: models.TealBlock, *, level: int) -> None:
         modified = peephole(block, level) or modified
     if level == 0:
         return
+    # do shuffling before dup insertion
+    if level > 1:
```
Argimirodelpozo: Nit: level > 1 here vs. level >= 2 a couple lines below (unless this is intentional for remembering to consider it for O1 too in the future)

src/puya/teal/optimize/move_shuffled_ops.py
daniel-makerx: Oh most definitely, thanks for spotting this!

I have fixed this, and turned the intc/bytesc checks into errors since they would also need to consider callsub if we started introducing constant blocks earlier in the pipeline.

src/puya/teal/optimize/main.py
daniel-makerx: I had >= 1 originally, but decided to be more conservative and do only O2 to start with, and so just dropped the `=`, but being consistent is better for sure.

src/puya/teal/optimize/move_shuffled_ops.py
```diff
@@ -0,0 +1,301 @@
...
+        window=slice(producer_idx, idx + 1),
+        new_ops=[*ops[expr_idx:idx], *producer_ops],
+        next_idx=producer_idx + 1,
+    )
+
+
+def _get_rotate_depth[T: (models.Cover, models.Uncover)](
+    op: models.TealOp, *, rotation_op_type: type[T]
+) -> int | None:
+    if is_stack_swap(op):
```
Argimirodelpozo, line 199: one last nit but maybe `is_stack_swap` is a good candidate to move to `_util`? just to avoid the dependency on peephole

daniel-makerx: Done

Argimirodelpozo (approved):
Looks great to me! Still would like for it to be double-reviewed by @iglosiggio before merging b.c. of it being a TEAL layer change.

src/puya/teal/optimize/move_shuffled_ops.py
```diff
@@ -0,0 +1,300 @@
...
+    <expr ΔN> is a series of ops with a stack delta of N
+    <producer_chain> is a MOVABLE_OP followed by zero or more PURE_TRANSFORM_OPS
+    """
+    ops = block.ops
+    modified = False
+    idx = 0
+    while idx < len(ops):
+        plan = _plan_drop_cover(ops, idx)
+        if plan is None:
+            plan = _plan_drop_uncover(ops, idx)
```
iglosiggio, line 97: nit:
```suggestion
        plan = _plan_drop_cover(ops, idx) or _plan_drop_uncover(ops, idx)
```

maybe? feels a bit cleaner

src/puya/teal/optimize/move_shuffled_ops.py
```diff
@@ -0,0 +1,300 @@
...
+        if plan is None:
+            idx += 1
+        else:
+            preserve_stack_manipulations(ops, plan.window, plan.new_ops)
+            idx = plan.next_idx
+            modified = True
```
iglosiggio, line 103: Then you can invert the condition here:

```suggestion
        if plan:
            preserve_stack_manipulations(ops, plan.window, plan.new_ops)
            idx = plan.next_idx
            modified = True
        else:
            idx += 1
```

src/puya/teal/optimize/move_shuffled_ops.py
```diff
@@ -0,0 +1,300 @@
...
+        )
+    if moving_op_code in _ITXN_RESULT_OPS:
+        return is_callsub or barrier_op_code == "itxn_submit"
+
+    if moving_op_code in _INTC_OPS | _BYTEC_OPS:
+        # barrier code assumes constant blocks are not present yet
+        raise InternalError(
+            "unexpected constant block reference op", maybe_barrier_op.source_location
+        )
+    return False
```
iglosiggio, line 277: I feel like we should to an allowlist instead. Whenever a new instruction is added it should be disallowed by default until we know it cannot cause issues :)

src/puya/teal/optimize/move_shuffled_ops.py
daniel-makerx: This version is cleaner, but I explicitly had it across two statements to ensure coverage reports would cover each branch separately.

You might see similar cases through out the code base where there are expressions with side effects or defaults where it is useful to see if it is used or not.

src/puya/teal/optimize/move_shuffled_ops.py
daniel-makerx: Another stylistic choice here (and through out the code base) is to have shorter branches first, the idea being when reading you can understand the simplest case first, before moving on the the more complex one.

src/puya/teal/optimize/move_shuffled_ops.py
daniel-makerx: I agree an allow list would be a safer choice here, possibly defined in `puya.teal.optimize._data` as it could be useful elsewhere?

PR #735, opened 2026-05-21
Fix: match branch bug

Argimirodelpozo:
This PR solves a compilation bug where a `match` statement could execute the wrong branch when two case values folded to the same constant during optimization. The `IRMutator` switch visitor overwrote colliding case keys, so a later branch could silently replace an earlier one, violating Python semantics.
In the proposed fix, the visitor for `match` in `IRMutator` now rebuilds the match value->block dict. in place, preserving case textual order. In this way unreachable branches are discarded in favor of earlier visited ones.

EDIT: will change the name from "switch" to match tomorrow 🙃 but its otherwise ready for review


achidlow:
So [when building the IR, we error](https://github.com/algorandfoundation/puya/blob/96e5c4af2c887c456f2c1c79f22522a29f8cb902/src/puya/ir/builder/flow_control.py#L57) if there are duplicate case values. I'm thinking maybe here we should do the same? It shouldn't matter if the duplicate is detected immediately or after optimisation I think.

Argimirodelpozo:
> So [when building the IR, we error](https://github.com/algorandfoundation/puya/blob/96e5c4af2c887c456f2c1c79f22522a29f8cb902/src/puya/ir/builder/flow_control.py#L57) if there are duplicate case values. I'm thinking maybe here we should do the same? It shouldn't matter if the duplicate is detected immediately or after optimisation I think.

I didn't check but wrongfuly assumed same semantic behavior than python. I agree it'd be best to keep it the same regardless on if it's product of an optimization or not.
Changed it to error on collision now 👍

achidlow (approved):

PR #737, opened 2026-05-26
Improve MIR x-stack scheduling

daniel-makerx:
Improves existing x-stack scheduling algorithm via a few mechanisms

* Allowing an x-stack ordering that is not constrained by the longest common sub-sequence of variable stores/loads
* Allowing variables that transit a block without being read and/or stored
* Allowing variables that are live on all in_blocks but one
* Allowing an x-stack variable to be copied instead of just moved 
* Considering the lifetime and usage of a variable when ranking the x-stack


src/puya/mir/stack_allocation/x_stack.py
```diff
@@ -1,364 +1,490 @@
...
+            local_id_loads=frozenset(local_id_loads),
+            live_in=frozenset(vla.get_live_in_variables(block.ops[0])),
+            live_out=frozenset(vla.get_live_out_variables(block.ops[-1])),
+        )
+    records = list(block_name_records.values())
 
-def get_x_stack_load_ops(record: BlockRecord) -> set[mir.AbstractLoad]:
-    block = record.block
-    assert block.x_stack_in is not None
+    # given blocks 1,2,3,4,5,6 and 7
```
Argimirodelpozo:
```suggestion
    # given blocks 1,2,3,4,5,6,7 and 8
```

daniel-makerx: Ended up going with `blocks 1-8`

src/puya/mir/stack_allocation/x_stack.py
```diff
@@ -1,364 +1,592 @@
...
+            if not self.is_multi_use(local_id)
+            and self.transit_cost(local_id) >= _MAX_TRANSIT_BLOCKS
         )
 
+    def ranked_locals(self) -> Sequence[str]:
+        """
+        Order x_stack_candidates across the group so locals stored later in predecessors
+        and loaded earlier in successors land shallower, with transient locals near the top.
+        """
+        # ollect per-local observations
```
Argimirodelpozo:
```suggestion
        # collect per-local observations
```

src/puya/mir/stack_allocation/x_stack.py
```diff
@@ -1,364 +1,592 @@
...
-            *(b.live_in for b in in_blocks),
-        ]
 
-        # determine shared l-stack variables for this edge_set
-        # determine all valid x-stacks (ordered)
-        first, *others = l_stacks
-        common_locals = frozenset(first).intersection(*others)
+# live-through locals are demoted when they would transit at least this many penalty blocks
+# current threshold chosen empirically via poe size_diff
+_MAX_TRANSIT_BLOCKS = 3
```
Argimirodelpozo: Just curious: how impactful was this parameter on the result across the full corpus? I'd assume small, hand tried a couple variants across a handful of contracts and changed like 4 bytes total on 4 and then plateau'd.

Argimirodelpozo (approved):
Looks good to me. The avg. for reconciling contradicting ordering across an edge set group is a _great_ idea

src/puya/mir/stack_allocation/x_stack.py
daniel-makerx: I just remeasured it and it has shifted slightly, probably because last time it was measured was before the ranking was refined. I'll update it to 2. 

Of note 4 is better for O2 but worse for O1 so there is probably some future work here to pick a better value using something other than a constant heuristic

| _MAX_TRANSIT_BLOCKS | O1 bytes | O2 bytes | O1 ops | O2 ops |
|---------------------|----------|----------|--------|--------|
| 0                   | 0        | 0        | 0      | 0      |
| 1                   | −103     | −130     | −58    | −67    |
| 2                   | −153     | −227     | −94    | −118   |
| 3                   | −128     | −216     | −78    | −111   |
| 4                   | −131     | −246     | −76    | −128   |
| 5                   | −131     | −246     | −76    | −128   |

iglosiggio (approved):
Omg, this was so difficult to review :exhausted:. Changes look great, we may have some possibility of improvement on `transit_mismatches` and the general local ranking procedure.

daniel-makerx:
> Omg, this was so difficult to review :exhausted:. 

Sorry! 🙂

> we may have some possibility of improvement on `transit_mismatches` and the general local ranking procedure.

Yeah there is still some potential there, a more varied corpus or at least more real world contracts might help to tune things better, and an approach that measures the cost more accurately could allow picking better orderings for each edge set

iglosiggio:
> > Omg, this was so difficult to review :exhausted:.
> 
> Sorry! 🙂

No problemo, thank YOU for the work!

> > we may have some possibility of improvement on `transit_mismatches` and the general local ranking procedure.
> 
> Yeah there is still some potential there, a more varied corpus or at least more real world contracts might help to tune things better, and an approach that measures the cost more accurately could allow picking better orderings for each edge set

Yup, I was really worried that the implementation could be fragile but one I understood how it worked I ended up pretty sure that the implementation is either good or at least possible miscompilations would be catched by our own assertions. I made some cosmetic changes rearranging the code a bit and bounced those to @Argimirodelpozo so we can doublecheck them before the final merge (prolly tomorrow)

PR #738, opened 2026-05-27
experimental: AWST References

iglosiggio:
## Proposed Changes

This PR adds a way to compress AWST JSONs. JSON nodes that carrying unique identifiers can be identified and referenced by other JSON nodes. This does not reduce node duplication during execution since `cattrs` does not deduplicate nodes based on identity information.


src/puya/awst/serialize.py
```diff
@@ -91,6 +152,8 @@ def structure_switch_cases(value: object, _: type) -> object:
     @functools.wraps(structure_method)
     def wrapped_structure[T](obj: object, cl: type[T]) -> T:
```
daniel-makerx, line 155: Is it worth having structuring produce ref and id fields too? That way the cattrs test cases will exercise this code and can serve has a reference implementation for other front ends?

iglosiggio: You mean structuring or destructuring? I think it would be a good idea to have ref and id fields for AWSTs serialized by puyapy. I don't know how we would implement it.

I spent almost a week trying to make structuring work via regular cattrs hooks and failed to do it. I imagine the destructuring path would have similar issues 😓.

daniel-makerx: Yes sorry I meant to say unstructuring. 

Another option might be to just add a new `test_cases_awst` test case that has the REF/ID fields from puya-ts, so at least there is some coverage in the puya test suite, perhaps using AWST from the reproduction of the original issue.

iglosiggio: That sounds like a good idea. I've talked w/@Argimirodelpozo about eventually having a drastic representation change (some kind of explicit type references + a type catalog) so this change feels kinda like a temporary improvement for `puya-ts` to use.

iglosiggio: Done :)

test_cases_awst/serialized_references/module.awst.json
```diff
@@ -0,0 +1,506 @@
+[
+  {
+    "_$%!#ID": 0,
+    "_type": "Contract",
+    "source_location": {
+      "file": "/contract.algo.ts",
```
daniel-makerx: You will need to use the `%DIR%` token (see the other module.awst.json files for examples) so the absolute paths are portable across CI / local development

iglosiggio: Thx!

Argimirodelpozo (approved):

PR #740, opened 2026-05-29
 feat: add optimisation that moves single use pure intrinsics to just before their consumer

daniel-makerx:
Offsets some of the op count regressions introduced in #739 that aren't covered by #733


achidlow (approved):

iglosiggio (dismissed):

iglosiggio:
Changes look mostly good but I think we should be even more careful than this. Not all `txn` values are constante across a single bblock (`log` can change the value of `txn NumLogs`, for example)

iglosiggio:
```python
import typing

from algopy import (
    Contract,
    log,
    op
)

class Reference(Contract):
        zero = op.Txn.num_logs
        log("test")
        log(op.Txn.num_logs)
        log(zero)
        return True

        return True
```

Ends up generating:
```
#pragma version 11
#pragma typetrack false

// repro.Reference.approval_program() -> uint64:
main:
    // repro.py:12
    // log("test")
    pushbytes "test"
    log
    // repro.py:13
    // log(op.Txn.num_logs)
    txn NumLogs
    itob
    log
    // repro.py:11
    // zero = op.Txn.num_logs
    txn NumLogs
    // repro.py:14
    // log(zero)
    itob
    log
    // repro.py:15
    // return True
    pushint 1
    return
```

daniel-makerx:
> Changes look mostly good but I think we should be even more careful than this. Not all `txn` values are constante across a single bblock (`log` can change the value of `txn NumLogs`, for example)

That is an excellent point, I think we had always considered all `txn` fields to be constant for a transaction but `txn NumLogs` is definitely an exception, similarly with the `gtxn` family (since we can't know the group index of a txn at compile time). This would apply for #733 too. Simplest for now would be to exclude that field from being moved

src/puya/ir/optimize/sink_single_use_intrinsics.py
```diff
@@ -0,0 +1,217 @@
...
+from puya.ir import models
+from puya.ir.avm_ops import AVMOp
+from puya.ir.optimize._utils import SSAReadTracker
+
+logger = log.get_logger(__name__)
+
+
+# "conservative" sets of fields that can't fail/change
+# and are therefore safely movable by this pass
+_MOVABLE_TXN_FIELDS: typing.Final = frozenset(
```
iglosiggio: This is OK for now but we should eventually have this info at the op level (and extract it from either langspec or something like that).

src/puya/ir/optimize/sink_single_use_intrinsics.py
```diff
@@ -0,0 +1,217 @@
...
+        "LocalNumByteSlice",
+        "ExtraProgramPages",
+        "Nonparticipation",
+        "StateProofPK",
+        "NumApprovalProgramPages",
+        "NumClearStateProgramPages",
+        "RejectVersion",
+    }
+)
+_MOVABLE_GLOBAL_FIELDS: typing.Final = frozenset(
```
iglosiggio: Same here.

src/puya/ir/optimize/sink_single_use_intrinsics.py
```diff
@@ -0,0 +1,217 @@
...
+    if not consumer_assignment:
+        return False
+
+    for block in subroutine.body:
+        new_ops = list[models.Op]()
+        for op in block.ops:
+            if (assignment_to_sink := consumer_assignment.pop(op, None)) is not None:
+                new_ops.append(assignment_to_sink)
+            new_ops.append(op)
+        block.ops[:] = new_ops
```
iglosiggio: isn't this a really inefficient way to do it?

We want to delay storing "edits" and applying them later. This idiom appears multiple times on our code and is always on this "copy everything bit by bit and replace" shape.

iglosiggio (approved):
Changes look ok. We can improve some stuff but LGTM instead of keeping it on the review queue :)

src/puya/ir/optimize/sink_single_use_intrinsics.py
achidlow: @iglosiggio generally list operations like this are fast - we’re not copying the full objects, just refs, so it’s like building a new list of pointers. It’s also not recommended generally to mutate a list as you go - so the alternative is copying the list just to iterate it. But then you also have to manage indexes etc. This code of just building a new list with what you want is something we found easier to read. It also depends on the frequency or likelihood of the operation - like in the code above this, we use list.remove which is O(N), because it’s less likely. (That’s my guess anyway, it was a little while ago, it may be the O(N) search is not justified and should just rebuild the list instead there too?). 

What is slow though is validation, which is why we assign by slice at the end here, and then do a full validation at the end of each optimisation pass (and why the first thing we’d generally do when debugging issues with that was to indent it so it runs after each function within a pass). 

All of the above may not apply here as strongly as it does in some other cases, but since you asked about the pattern, the main things are that yeah, shallow list copies are generally cheap to build,  easy to reason about, and if you’re iterating you should be copying anyway if you need to modify, and also that the slice assignment is “load bearing” :P

PR #741, opened 2026-06-03
Fix: routing bug when overriding approval program

Argimirodelpozo:
Fixes an arc4 routing bug. 
When overriding an approval program, and in the presence of both non-zero and zero (i.e. no-op) OCA methods, routing for a method call with non-zero OCA would still fall through a no-op check and thus fail.


iglosiggio (approved):

iglosiggio:
LGTM

PR #743, opened 2026-07-01
feat: AVM variable length branches in Puya assemble

Argimirodelpozo:
This PR introduces varint zig-zag encoding for single target jumps (`b`, `bnz`, `bz` and `callsub`) in Puya's assemble, matching go-algorand's AVM v13+ behavior (see [this PR](https://github.com/algorand/go-algorand/pull/6600)).

Tests mostly mirror those in go-algorand. When AVM 13 is available in localnet, they should be refactored into parity checks with algod (same as all pre-existing test programs and examples).
For the same reason, we temporarily exclude existing v13 programs from being parity checked against algod.


src/puya/ussemble/assemble.py
```diff
@@ -290,9 +308,79 @@ def _resolve_template_vars[T: (int, bytes)](
...
+                    op.source_location,
+                )
+            # TODO: for now we don't have any ops that are more than 1 byte,
+            # much less a branching op. But a soon to be merged PR will carry
+            # multi-byte opcodes. So probably a good idea to make these magic
+            # +/-1s to be the opcode size when that's done
+            if needed < op_sizes[index] - 1:
+                op_sizes[index] = needed + 1
+                changed = True
+    return pcs, branch_offsets
```
iglosiggio, line 375: This should be equivalent to https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/assembler.go#L2521-L2564 right?

iglosiggio (approved):
Changes look OK, will probably have to take second look at `_compute_pcs` later.

Should we migrate most of our tests to v13 once the new algod is out?

src/puya/ussemble/assemble.py
```diff
@@ -290,9 +308,79 @@ def _resolve_template_vars[T: (int, bytes)](
...
+            )  # add 1 b.c. of opcode size (see TODO below)
+        else:
+            op_size = len(
+                _encode_op(
+                    op, get_label_offset=lambda _: 0, varint_label_immediates=varint_branches
+                )
+            )
+            assert op_size, "expected non empty bytecode"
+            op_sizes.append(op_size)
+
```
iglosiggio, line 336: All that happens after this only makes sense if `varint_branches` is `True`. Should we have an early exit here?

src/puya/ussemble/assemble.py
Argimirodelpozo: Yes! It has since moved but I assume you meant `findBranchSizes`, permalinked for reference: https://github.com/algorand/go-algorand/blob/9c3e95b4b6df71aafe850024c1dc8f6b26b652f8/data/transactions/logic/assembler.go#L2601

src/puya/ussemble/assemble.py
Argimirodelpozo: Yeah I think it makes sense.
It still has to compute pcs at least once, but I agree that it improves readability to add the early exit.

iglosiggio (approved):

PR #744, opened 2026-07-11
feat: AVM 13 off-curve salting and `#pragma autosalt`

Argimirodelpozo:
This PR introduces support for off-curve autosalting for programs, a significant AVM 13 change (see go-algorand PRs [6592](https://github.com/algorand/go-algorand/pull/6592) and [6653](https://github.com/algorand/go-algorand/pull/6653)).

In the context of TEAL programs, salting means appending an `intcblock 1 <salt>` to the bytecode only when the unsalted program hash (`sha512_256(b"Program" || bytecode)`) decodes to an `ed25519` point; a program that already hashes off-curve is always left unchanged.

Like in go-algorand, the check is more lenient than what usually provided crypto library implementations tends to be, and so the implementation of the check is vendored.

It is worth mentioning that, since Puya has a declaration of intent with regards to mode when it comes to programs (Contract or LogicSig), we omit the stateful/stateless check. Instead, the emitted `#pragma autosalt` records the resolved setting so the produced TEAL output produces the same bytecode whether assembled by our own bundled assemble or algod's (being default `false` for contracts and `true` for lsigs; this kind of salting makes little sense for contracts but we still allow devs to use it if they so please -warning them just in case-). Note that below AVM v13, algod does not auto-salt, so a resolved-off program omits the redundant `#pragma autosalt false` in this specific case.

The behavior matrices for lsigs and contracts are as follows.

## Logic signatures

Salted by default on **any** AVM version (we are stricter than go-algorand's default here, which
only auto-salts v13+; the behavior is expressed to algod via the pragma in order to maintain parity).

| AVM version | `autosalt` | Behavior         | Pragma emitted | Warning |
| ----------- | ---------- | ---------------- | -------------- | ------- |
| v12- | unset   | Salt if on-curve | `true`  | No  |
| v12- | `True`  | Salt if on-curve | `true`  | No  |
| v12- | `False` | No autosalt      | *none*  | Yes¹ |
| v13+ | unset   | Salt if on-curve | `true`  | No  |
| v13+ | `True`  | Salt if on-curve | `true`  | No  |
| v13+ | `False` | No autosalt      | `false` | Yes¹ |

¹ `autosalt=False leaves this LogicSig's address on-curve, i.e. potentially decodable as a public key`

## Contracts

Never salted by default, since a contract's program hash is not used as a spendable address.
Applies identically to the approval and clear-state programs.

| AVM version | `autosalt` | Behavior         | Pragma emitted | Warning |
| ----------- | ---------- | ---------------- | -------------- | ------- |
| v12- | unset   | No autosalt      | *none*  | No  |
| v12- | `True`  | Salt if on-curve | `true`  | Yes² |
| v12- | `False` | No autosalt      | *none*  | No  |
| v13+ | unset   | No autosalt      | `false` | No  |
| v13+ | `True`  | Salt if on-curve | `true`  | Yes² |
| v13+ | `False` | No autosalt      | `false` | No  |

² `autosalt=True has no security benefit for a contract since its program hash is never used as a spendable address`


src/puya/ussemble/assemble.py
```diff
@@ -140,15 +145,45 @@ def get_label_offset(label: models.Label) -> int:
...
-        bytecode=b"".join(bytecode),
+        bytecode=final_bytecode,
         debug_info=build_debug_info(ctx, pc_ops, pc_events),
         template_variables={
             var: ctx.provided_template_variables.get(var, (None, None))[0]
             for var in ctx.template_variable_types
         },
         stats=_get_op_stats(op_stats),
         instruction_boundaries=pcs,
+        salt=salt,
```
iglosiggio, line 165: To be honest I find it weird to extract it this way. I would prefer a salted/nonsalted flag and you know that salted means the last byte is the salt.

Still, this is probably more legible (and we can change it later, it is just a dumb impl detail).

src/puya/utils.py
```diff
@@ -130,6 +130,45 @@ def method_selector_hash(method_signature: str) -> bytes:
...
+    p = _ED25519_P
+    # The low 255 bits are y; the top bit encodes the sign of x, which does
+    # not affect whether a point exists.
+    y = (int.from_bytes(public_key, "little") & ((1 << 255) - 1)) % p
+    u = (y * y - 1) % p
+    v = (_ED25519_D * y * y + 1) % p
+    # A point exists iff x^2 = u / v has a solution, i.e. u / v is a square.
+    x = (u * pow(v, 3, p) * pow(u * pow(v, 7, p) % p, (p - 5) // 8, p)) % p
+    vxx = (v * x * x) % p
+    return vxx == u % p or vxx == (-u) % p
```
iglosiggio, line 169: Please wire this to @joe-p curvecheck repo (https://github.com/joe-p/curvecheck) so we can cross-validate w/the other impls.

test_cases/autosalt/out/module.awst
```diff
@@ -0,0 +1,209 @@
...
+  baremethod test_cases.autosalt.contract.UnsaltedContractV12.__algopy_default_create(): void
+  {
+  }
+  
+  subroutine algopy._contract.Contract.__init__(): void
+  {
+  }
+}
+
+contract SaltedContractV12
```
iglosiggio, line 181: Should we add some decoration for auto salt at the awst printer?

It can go on a later PR but I think the answer is yes.

tests/test_assemble.py
```diff
@@ -35,6 +35,9 @@ def test_assemble_matches_algod(
         )
         if int(program.teal_src.splitlines()[0].split()[2]) >= 13:
             continue
+        # TODO: delete this once localnet supports `#pragma autosalt`
+        if "#pragma autosalt" in program.teal_src:
+            continue
```
iglosiggio: We should do this now right?

iglosiggio (approved):
Looks pretty good, some small comments on remaining stuff. The last one (skipping assembly tests) is the only blocker IMO.

tests/test_assemble.py
Argimirodelpozo: Oh yeah! And also got rid of the general AVM 13 exclusion right above. 
Feels nice to finally delete these

src/puya/utils.py
Argimirodelpozo: It is an exact copypaste of the one in py-algorand-sdk (only change is the constants are added above the function whereas the sdk impl. already had them available) so I'm not worried it's undertested. However I agree it'd be good to add it, I'll PR it over there later

test_cases/autosalt/out/module.awst
Argimirodelpozo: Yeah I agree it'd be good to show it. We don't have decorators for most of the the other contract/lisg options, so I'll do it in a separate small PR and add all of them along with autosalt.

iglosiggio (approved):

PR #745, opened 2026-07-13
feat: logic signature group transaction args.

Argimirodelpozo:
This PR implements group transaction references as typed arguments in logic signatures.
They work exactly the same as in application ABI methods: they are taken in order of appearance in the signature and represent a reference to a prior transaction in the group (according to their order of appearance in the function signature). If they are typed, they also have their type validated in the same fashion (e.g. if they are a `gtxn.PaymentTransaction` then their `TypeEnum` field is asserted to be 1 for `payment`).
We also update outdated docs. to reflect support for typed args. in logicsigs.


iglosiggio (approved):
lgtm

PR #748, opened 2026-07-22
feat: support for new AVM 13 ops. and fields

Argimirodelpozo:
Changes to support new ops. and fields in AVM 13.

Most of the changes here are autogenerated from the consumed langspec.
However, since the langspec. was updated and augmented upstream, we had to update our scripts as well.

Also, this PR introduces for the first time opcodes of more than 1 byte. We model them with a `sub_code` field.

Finally we add compilation tests for all new ops. and op fields.

_Note that the aforementioned langspec. changes need to be merged in go-algorand before we merge this PR_


iglosiggio (approved):
LGTM, we should doublecheck the run mode stuff since I added extra info on all operations removing the "dunno" enum value :)

Argimirodelpozo:
> LGTM, we should doublecheck the run mode stuff since I added extra info on all operations removing the "dunno" enum value :)

I took another look and the only run mode discrepancy is `itxn_field` which is actually an improvement over the old `any` (i.e. no `itxn_next` may be used in an lsig, and since `itxn_begin`, `itxn_next` and `itxn_submit` were explicitly app only already and there is no use for next outside that context it makes a lot of sense). So it's good to go!

PR #749, opened 2026-08-07
feat: increase maximum program pages

Argimirodelpozo:
Changes to support the new extra program pages in `CompiledContract`s.
Basically the tuple representing program pages should add a couple extra `Bytes` elements so it fits the new maximum program size.


tests/analyse/reti.awst.json.gz
iglosiggio, line 1: What happened here?

iglosiggio (approved):
Changes look good. I don't understand why the reti test got deleted.

tests/analyse/reti.awst.json.gz
Argimirodelpozo: Same changes as the other AWST file. Didn't get deleted just decompressed, changed (i.e. all compiled contract tuples widened) then recompressed. I think the size being reduced is because of --best on gz. I think it's harmless but we could go back to the lower level, don't feel strongly about it.

PR #750, opened 2026-08-23
fix: bug in repeated loads elimination (multiple writes then a read to the same key)

Argimirodelpozo:
Fixed an iteration bug in repeated loads elimination causing a third write in a row to the same key to be skipped, instead getting the second one cached for a subsequent read in the same block.

_Note: regression in size from `conditional_execution` is because it just so happened that some of the skipped writes would prevent the full cache cleanup. A future PR could improve the "naive" full cache flush to something less conservative._


joe-p (approved):
One question: Why are there diffs for the conditional_execution output? 

If that is expected, LGTM

Argimirodelpozo:
> One question: Why are there diffs for the conditional_execution output?

The diffs are mostly a coincidence. With loads elim. we take a naive approach and clean the cached values w/any read or write. The skipped values just happened to not be relevant and therefore would prevent the cache from being cleared and things would end up in such a way that the reads could be properly eliminated. But its fortuitous as you are incorrectly blindly skipping an op here and there.
We could improve the strategy for the elims. later.

PR #751, opened 2026-09-14
fix: write lengths for (bitpacked) bool array in box + OOB bitpacked access

Argimirodelpozo:
This PR fixes a bug in bitpacked representation for bitpacked arrays inside boxes (whether by themselves or inside a containing struct).
Pushes to a packed bool array now correctly push 0 or 1 extra bytes according to the alignment of the added bit, plus updating offsets accordingly. Access off limits (i.e. to padding bits) will now fail.


src/_puya_lib/box_arc4.py
```diff
@@ -103,3 +103,100 @@ def _box_extract_u16(box_key: Bytes, offset: UInt64) -> UInt64:
...
+    while write_offset < new_arr_len:
+        byte_offset = data_offset + write_offset // 8
+        bit_offset = write_offset % 8
+        byte = op.Box.extract(box_key, byte_offset, 1)
+        while bit_offset < 8 and write_offset < new_arr_len:
+            byte = op.setbit_bytes(byte, bit_offset, op.getbit(new_items_bytes, read_offset))
+            bit_offset += 1
+            write_offset += 1
+            read_offset += read_step
+        op.Box.replace(box_key, byte_offset, byte)
```
iglosiggio: LGTM but add a ticket somewhere to optimize this. It feels pretty inefficient to go bit-by-bit (also take modulus multiple times and all that).

iglosiggio (approved):

src/_puya_lib/box_arc4.py
Argimirodelpozo: Absolutely, there is room for improvement here. I mostly copied the existing patterns in dynamic array so they behave exactly the same and we get the bug out of the way.

